PR #119, opened 2022-09-20
Deleted dependency section

zsmith27:
Probably not needed because it managed by the description file in the package.


cristinamullin:
Thanks for the demo today Zach! I agree, we can take this out.

cristinamullin (approved):

PR #121, opened 2022-09-20
ignore_attributes in readWQPdata

ldecicco-USGS:
I've added a GitHub Action yaml, and then did my best to get rid of notes/warnings. Ideally you'll want to make sure each pull request doesn't add any new notes/warnings/errors, so this GH Action check could do that automatically. 

If you want to speed up the checks, you can do fewer (I can put an in-line note where I'm talking about). If you want to automatically deploy to a github page, you could try using something like this:
https://github.com/USGS-R/dataRetrieval/blob/main/.github/workflows/pkgdown.yaml

ANYWAY, 
The point I wanted to make with this PR is just to add `ignore_attributes=TRUE` to any `readWQPdata` calls. Normally, if you did something like this:

```r
df <- readWQPdata(siteid = c("USGS-054064785",
                                        "USGS-430305089260600"),
                             characteristicName = "Phosphorus")
```
`dataRetrieval` will go out and get the data, but ALSO go get the site information via `whatWQPsites`. You can access that via the "siteInfo" attribute:
```r
siteInfo <- attr(df, "siteInfo")
nrow(siteInfo)
[1] 2
```
Since you are doing that independently, and joining that site information directly to the data, you can potentially save a lot of time by dropping the attributes (and therefore not calling `whatWQPsites` twice):

```r
df2 <- readWQPdata(siteid = c("USGS-054064785",
                                        "USGS-430305089260600"),
                                 characteristicName = "Phosphorus",
                                 ignore_attributes = TRUE)
siteInfo2 <- attr(df2, "siteInfo")
nrow(siteInfo2)
NULL
```

Alternatively you could keep the attributes, and drop the lines like this:
```r
sites.DR <- dataRetrieval::whatWQPsites(WQPquery)
```
instead:
```r
sites.DR <- attr(WQPquery, "siteInfo")
```


.github/workflows/R-CMD-check.yaml
```diff
@@ -0,0 +1,49 @@
...
+
+    name: ${{ matrix.config.os }} (${{ matrix.config.r }})
+
+    strategy:
+      fail-fast: false
+      matrix:
+        config:
+          - {os: macOS-latest,   r: 'release'}
+          - {os: windows-latest, r: 'release'}
+          - {os: ubuntu-latest,   r: 'devel', http-user-agent: 'release'}
```
ldecicco-USGS, line 23: You could drop line 23-25 to speed up the checks, then only macs and windows are being tested.

R/DataDiscoveryRetrieval.R
```diff
@@ -23,7 +23,12 @@
 #' 
 #' @export
 #'
-
+#' @examples
```
ldecicco-USGS, line 26: I'd highly recommend adding more examples to all the functions. These will get tested as the package is checked. Also, they get run and the output is displayed in the pkgdown output. I'm a huge fan of lots of examples.

R/GenerateRefTables.R
```diff
@@ -147,4 +147,146 @@ UpdateMeasureUnitRef <- function() {
   utils::write.csv(WQXunitRef, file = "inst/extdata/WQXunitRef.csv", row.names = FALSE)
 }
 
+#' Generate Unique Harmonization Reference Table
```
ldecicco-USGS, line 150: This came when I did an upstream pul...not sure why it's in this pull request....

R/Transformations.R
```diff
@@ -392,7 +392,7 @@ HarmonizeData <- function(.data, ref, transform = TRUE, flag = TRUE) {
   ) %in% colnames(.data)) == TRUE) {
 
     # if class(ResultMeasureValue) != numeric, run special char function
-    if (class(.data$ResultMeasureValue) != "numeric") {
+    if (!is.numeric(.data$ResultMeasureValue) ) {
```
ldecicco-USGS, line 395: These are relatively new NOTES in R. I think they started in ~4.1 or so. Anyway, it's always good to remove as many notes as possible.

vignettes/WQPDataHarmonization.Rmd
```diff
@@ -85,7 +85,7 @@ CharacteristicName
 ActivityMediaName
 endDate
 
-Tip 1) All the query filters for the WQP work as an AND but within the fields there are ORs. So for example, characteristics – if you choose pH & DO – it’s an OR. Similarly, if you choose VA and IL, it’s an OR. But the combo of fields are ANDs, such as State/VA AND Characteristic/DO". "Characteristic" and "Characteristic Type" on the WQP also work as an AND.
+Tip 1) All the query filters for the WQP work as an AND but within the fields there are ORs. So for example, characteristics - if you choose pH & DO - it's an OR. Similarly, if you choose VA and IL, it's an OR. But the combo of fields are ANDs, such as State/VA AND Characteristic/DO". "Characteristic" and "Characteristic Type" on the WQP also work as an AND.
```
ldecicco-USGS, line 88: There were some non-ASCII characters in here. That usually happens when copy/pasting from Word.

tests/testthat/test-DataDiscoveryRetrieval.R
```diff
@@ -1,3 +1,114 @@
 test_that("multiplication works", {
   expect_equal(2 * 2, 4)
 })
+
+test_that("TADAdataRetrieval ", {
+  tada1 <- TADAdataRetrieval(siteid = c("USGS-054064785",
+                                        "USGS-430305089260600"),
+                             characteristicName = "Phosphorus")
+  # you could just pick the important columns:
+  expect_true(all(c("OrganizationIdentifier",
```
ldecicco-USGS, line 10: You can make a lot more tests on the output here, I just am checking that all the columns we expect are there.

ldecicco-USGS:
Ideally you'd be able to set this repo up that the GitHub Actions can automatically run without any review. You can check it from my fork:
https://github.com/ldecicco-USGS/TADA/actions/runs/3093018055
To see that all the checks passed, but ideally you'd see that in this pull-request stream.

R/GenerateRefTables.R
cristinamullin: I moved this to the Transformations.R file recently because it makes more sense to have there, not sure why it is still here. I'll make sure to review & resolve this

R/DataDiscoveryRetrieval.R
cristinamullin: Great suggestion, do you recommend to always include \donttest for examples like this?

R/Transformations.R
cristinamullin: Thank you so much for your help with these notes! It's been on my list to address, but I wasn't sure how to do so.

cristinamullin (approved):

R/DataDiscoveryRetrieval.R
ldecicco-USGS: If you ever expect to push TADA to CRAN, you'll want to add `\donttest` to any example that makes an internet call. If your function is just doing something local, you don't need the donttest. Also, if you never plan to put it on CRAN then it's not a big deal either.

PR #127, opened 2022-10-01
Lazy load newest ref tables (fixes #72)

### LINKED ISSUE #72: Research how run functions upon loading package to R
We would like TADA to use the most up-to-date reference tables each time its used. Running the generate ref table functions automatically upon loading the package into R would allow us to do this. Do some research to learn how we can do that.

If it's not clear or there is no standard way to that, consider other approaches to achieving the goal of automatically refreshing tables

nx10:
Hi,

just a quick PR for hacktoberfest.

https://github.com/USEPA/TADA/issues/72 mentions the use of `.onLoad` and `.onAttach` to update the data system files:

This is very unusual behavior for an R package. Although I don't know of any official documentation mentioning this, but a package should probably not modify it's own files after installation. This PR realizes a different approach to solve this:

The (internal) functions `GetWQXCharValRef()` and `GetMeasureUnitRef()` are introduced to lazily fetch and return the newest tables on demand. This means that they will cache the tables after downloading once (per session) to improve performance. No data is downloaded when none of them are called. If the download is unavailable (e.g. no internet connection) the internal tables are returned (and the user is notified).

Functions `UpdateWQXCharValRef()` and `UpdateMeasureUnitRef()` are still available to update the static project files (I assume this is what you used them for).

Cheers,
Florian


cristinamullin (approved):

cristinamullin:
Hi Florian, thank you for taking on this issue and for the elegant solution! Great work

PR #136, opened 2022-10-07
Reduced install and load to bare minimum given dependencies in DESCRIPTION. 

jbousquin:
Installing TADA with dependencies ensures everything from the DESCRIPTION import list is installed. I tested the suggested (optional) installs and none of their functions are used directly in this vignette (loading them could conflict with standard functions by the same name so best to not). I tested a fresh install and it doesn't appear any of the suggested are being used indirectly either. I split the install and load originally because they were longer, now that it's short you could combine, but this makes it easier to not re-install if known to already be installed.

### DIFF
--- a/vignettes/WQPDataHarmonization.Rmd
+++ b/vignettes/WQPDataHarmonization.Rmd
@@ -33,7 +33,7 @@ or devtools. dataRetrieval will be downloaded from CRAN, but the development
 version can be downloaded directly from GitHub (un-comment).
 ```{r, results = 'hide', message = FALSE, warning = FALSE}
 # To install the latest version of TADA:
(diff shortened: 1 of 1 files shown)


jbousquin:
Unclear why this wasn't an issue before, but knitr isn't getting mirror from rstudio (see [stack](https://stackoverflow.com/questions/33969024/install-packages-fails-in-knitr-document-trying-to-use-cran-without-setting-a)), simple solution is to include it in install:
```R
install.packages("remotes", repos = "http://cran.us.r-project.org")
```

Could test for remotes before installing but there is a tradeoff between efficient and simple.

cristinamullin (approved):

PR #145, opened 2022-10-20
Test depth profile data

jbousquin:
Added those units tests - should all be passing (tested locally and via actions). (Note last 3 commits were just to test and revert)

Only concern is that the order of the rows seems to change during conversion (hard to tell since these are just values and units, but row 2 is converted and row 1 then has the result).

### DIFF
--- a/.github/workflows/R-CMD-check.yaml
+++ b/.github/workflows/R-CMD-check.yaml
@@ -2,9 +2,9 @@
 # Need help debugging build failures? Start at https://github.com/r-lib/actions#where-to-find-help
 on:
   push:
-    branches: [develop, test-DepthProfileData]
+    branches: [develop]
   pull_request:
-    branches: [develop, test-DepthProfileData]
+    branches: [develop]
 
 name: R-CMD-check
 
--- a/.github/workflows/R-CMD-check.yaml
+++ b/.github/workflows/R-CMD-check.yaml
@@ -2,9 +2,9 @@
 # Need help debugging build failures? Start at https://github.com/r-lib/actions#where-to-find-help
 on:
   push:
-    branches: [develop]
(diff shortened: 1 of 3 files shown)


cristinamullin:
Good catch. I also noticed the row order changed. I think that has to do with the column reordering, which is probably not needed anymore. My original thought when we designed this function was that users would want to see the added columns next to the relevant columns in the dataframe, and not at the end. But I'm reconsidering now since R users are used to new fields being added at the end, not in the middle of the dataframe. Also since the reordering is only relevant to transform=FALSE now, it might more confusing than helpful. What do you think is more user friendly?

jbousquin:
I'm seeing the rows get re-ordered before that, within the loop that merges to the conversion factors (~lines 405-426), probably via merge (possible solutions on [stack](https://stackoverflow.com/questions/17878048/merge-two-data-frames-while-keeping-the-original-row-order)). Once it is fixed the index in the last test just needs to be changed 1 -> 2. You prefer that get resolved here before merging the tests or we create an issue/PR for it?

As a general rule I don't re-name/re-order cols unless otherwise they seem like something they aren't or are really ugly. It can make it harder to follow what the code did and the end user may want cols to have different names/order for their workflow anyway (doing it multiple times is inefficient). Particularly here it adds a good chunk of code. A third option is to add an arg where users can specify order with new cols next to their origin or at the end. A topic for discussion/the working group?

cristinamullin:
Great idea - I added an issue for the discussion topic: https://github.com/USEPA/TADA/issues/147 
I also created a new label to easily tag for issues that would be good workgroup topics (TADAWorkingGroupTopic)

cristinamullin:
Thanks so much again for contributing the tests & for all your helpful contributions & reviews! Happy friday

PR #146, opened 2022-10-21
Cm depth

cristinamullin:


cristinamullin:
As part of this, I moved the depth function to transformations.R and changed the name to ConvertDepthUnits to be more intuitive. I added the test back in & believe this addresses the ordering issue too.

R/DataDiscoveryRetrieval.R
```diff
@@ -1,7 +1,7 @@
-#' Generate TADA-Compatible Dataset with WQP Data
+#' Generate TADA-Compatible Data with WQP Data
```
jbousquin, line 1: This is minutiae, but Generate TADA-Compatible _dataframe_ would be more consistent with the other replacements in this commit. Also consider _from_ or _using_ WQP Data, since that makes it clearer the WQP data is an input.

jbousquin:
Good updates - I like the more specific 'dataframe' vs dataset/data (made some minor comments, up to you to keep as is or change). I think where you've moved these two functions is more intuitive as well. Looks like there are just some errors in vignette and building the contributing file causing checks to fail.

R/DataDiscoveryRetrieval.R
cristinamullin: Agree & added a commit for this change

Issue #152, opened 2022-10-27
TADAProfile Check


cristinamullin:
Consider including a function to check for all the required/expected fields as part up data retrieval up front (optional for users to check up front - might be most useful for users brining in outside WQP data). However the current solution that checks for required columns within each function, might be most flexible. 

Some users might make adjustments to the data frame that could impact an individual functions ability to run. Also, functions can be run in any order so we would need to think about where in the work flow the encompassing TADA Profile column check makes most sense (if at all). 

For example, we are considering including a function that deletes all empty columns - however this could make it so other functions won't work or be relevant. For example - If we were to remove empty columns as part of the data cleaning process then it could impact the ConvertDepthUnit function which currently relies on at least one column for re-ordering of the columns that is not actually used in the function (I think I want to try to remove this dependency for this reason). It might be helpful to simply keep all columns and write a function that tests the entire data frame for all fields that will be required to run all TADA functions.

Another thought, if we put the column check as part of data retrieval and we can ask users them not to remove empty columns until the very end... that should work for most users. But we are also trying to design TADA so outside data can be used as well as long as it has the required fields to run all functions. These users will also need a way to test if all the columns needed are included. 

Maybe the best solution would be to create a stand alone function that checks for all the fields required to run all TADA functions (it would be optional)?

cristinamullin:
This would probably fit best in utilities. In utilities now, there is a draft TADAprofileCheck function that is not currently used anywhere) & needs updating.

What if the generalized check columns function could be done up front (as part of data retrieval or when a user supplies a dataframe). It could tell a user which functions they will be able to run using their input dataset. 

Additionally, it could be used as a check then within each individual function. 

For example:
CheckColumns (default to check all TADA functions; or user can specify specific functions to check)

returns notes e.g.,
-your dataframe includes columns needed to run the following TADA functions: list...
-your dataframe does not include the columns needed to run the following TADA functions: list...

non-default option where function user can specify specific functions to check (would use function name as input & check expected columns for that function name)
- check specific function(s), input as list of functions
- include required columns for each function here & test if they are present
This option could be used as the check within each TADA function as well?

cristinamullin:
"Should it have the flexibility to also confirm that those columns have results other than NaN (don't think so - do one thing well)"

I agree this should be a separate function. We have a bare bones function started in utilities now for this too: RemoveEmptyColumns

PR #154, opened 2022-10-31
Check functions

### LINKED ISSUE #151: Checks for cols in colnames 2x
In many of the functions in Transformations a stop error occurs if expected_cols in .data == FALSE, and then the main function occurs on the conditional that this is TRUE. I wouldn't expect this to change, is it adequate to just check the first time? If not could these be combined, i.e., if FALSE stop() else {continue function}?

jbousquin:
Resolves #151 and #152 by adding two small check functions in Utilities.R and implementing throughout. Error messages manipulated to mirror/pass tests for ConvertDepthUnits function, should be easier to standardize with these new functions rather than having a unique error string for each.


cristinamullin (approved):
Scanned these & they look great. I added a note to the TADA Coding Cheat Sheet for this: 

14.	Utilities.R: Within each new function, add utilities if needed
i.	checkColumns: check if required columns to run function are available
ii.	checkType: check if type is a dataframe, if logical (boolean)

PR #168, opened 2022-12-06
Retrievaledits cm

cristinamullin:


cristinamullin:
new retrieval function for joining profiles from the WQP. Added tests for new function and new test data folder. Also added the data folder and added some sample data.

PR #172, opened 2022-12-09
InvalidFraction example

cristinamullin:
Added example to InvalidFraction function in ResultFlagsDependent.R


katiehealy (commented):
Thanks for the example! Looks good to me.

katiehealy (approved):
Sorry - I guess I just commented instead of approved! Adding this comment to approve.

PR #175, opened 2022-12-22
Update ResultFlagsDependent.R

katiehealy:
Added an "errorsonly" parameter to InvalidFraction() which filters data to show only the invalid characteristic-fraction combinations when errorsonly = TRUE. **Question: should errorsonly filter to show only "Invalid" labeled combinations, or should it also include "Nonstandardized" combinations (where WQX.SampleFractionValidity is NA)?


katiehealy:
Checks failed for a few reasons: 
- example of HarmonizeData() function didn't have a "ref" input; however, the documentation states this would be an optional argument - function may need to be updated so that checkType(ref, "data.frame") only happens if a reference table is provided
- document() needs to be run to update documentation for functions with parameter "errorsonly" added

cristinamullin (approved):

Issue #177, opened 2022-12-29
InvalidCoordinates


cristinamullin:
Considering adding another option for the "clean_outsideUSA" function input. This new option would allow users to change the sign of the lat or long measurement so that the site is within the US. This is to deal with sites that are known to be in the US, but the site lat/long data was entered wrong. Data owners should go back and FIX this raw data through WQX... they can email the WQX helpdesk WQX@epa.gov for assistance in fixing their raw data in WQX/WQP.

If LAT < 0, change LAT to positive
If LongitudeMeasure > 0 & LongitudeMeasure < 145 , change LONG to negative

Tests and make sure the logic above works. 

There may be exceptions: 
- Guam has a neg latitude so we would need to account for that
- Guam also has pos long
- USMI (Mariana islands) pos lat and long

cristinamullin:
@katiehealy FYI I linked this issue to your pull request, so it will close automatically when that pull request is merged into develop

PR #178, opened 2022-12-29
option to change sign of lat/long outside USA

katiehealy:
Changed clean_outsideUSA argument in InvalidCoordinates() to accept three options: "no" (default), "remove", or "change sign" to address issue #177
Question: after changing the sign of coordinates outside the USA, should the flag from TADA.InvalidCoordinates column be removed? (I left it in, so it will still show up if errorsonly = TRUE.)

### DIFF
--- a/R/DataDiscoveryRetrieval.R
+++ b/R/DataDiscoveryRetrieval.R
@@ -113,6 +113,10 @@ TADAdataRetrieval <- function(statecode = "null",
   # Retrieve all 3 profiles
   results.DR <- dataRetrieval::readWQPdata(WQPquery,
                                            ignore_attributes = TRUE)
+  #check if any results are available
+  if ((nrow(results.DR) > 0) == FALSE) {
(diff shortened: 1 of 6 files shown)


cristinamullin:
This is still failing due to an update needed to the "checkType" function in the Utilities.R file (@jbousquin this wasn't an issue before but is now since we made this InvalidCoordinates a bit more complex). This function currently can only check for a single type (e.g. logical (boolean) or character). This must be updated to deal with the more complex situation like we have with the "InvalidCoordinates" function in ResultFlagsIndependent.R. This function may have both boolean and character input types when for example (see below):
TADAProfileClean1 <- InvalidCoordinates(TADAProfile, clean_outsideUSA = "no", clean_imprecise = FALSE)

The vignette error is: 
   Quitting from lines 312-317 (WQPDataHarmonization.Rmd) 
   Error: processing vignette 'WQPDataHarmonization.Rmd' failed with diagnostics:
   no must be of class 'logical'

cristinamullin (approved):

PR #187, opened 2023-01-20
Update DataDiscoveryRetrieval.R

ehinman:
Added "project name" to join because it is an overlapping column between join2 and projects.DR dataframes. Fixes issue where output has ProjectName.x and ProjectName.y columns.


cristinamullin (commented):
Great catch. Can you please make this same change to update the join function and TADABigDataRetrieval function as well in this .R file? For future discussion, we may want to leverage the join function in tatadataretreival and tadabigdataretreival in the future, so that we only need to update 1 join function and it will trickle down to these as well.

cristinamullin (approved):

ehinman:
@cristinamullin It doesn't look like the QAPP columns from project are added in the TADABigdataRetrieval. The project query is commented out and the join isn't there. Would you like me to change that part as well? Or, I could make improvements in the JoinWQPProfiles function and then embed that into the TADAdata- and TADABigdataRetrieval functions?

cristinamullin:
Yes, that would be great, thank you! I like the idea of making improvements in the JoinWQPProfiles function and then embeding that into the TADAdata- and TADABigdataRetrieval functions. What do you think?

ehinman:
I have made the requested changes and committed to the fix_project_join branch. One issue that I will investigate further: TADABigdataRetrieval does not return as many records as TADAdataRetrieval on the same query. Example:
`tada2 <- TADABigdataRetrieval(startDate = "2022-01-01", endDate = "2023-01-01", characteristicName = "Escherichia coli",
siteType = "Stream")`
and
`tada2a <- TADAdataRetrieval(startDate = "2022-01-01", endDate = "2023-01-01",characteristicName = "Escherichia coli", siteType = "Stream")`

tada2 returns 21354 records, while tada2a returns 31178 records.

ehinman:
Hi @cristinamullin! When you are ready, you can take a look at this pull request and merge it if appropriate. Thanks so much for your help and review.

cristinamullin (approved):

PR #193, opened 2023-01-30
TADAbigdataRet chunking by site

ehinman:
- Chunks by site list generated by readWQPsummary, rather than sites within each FIPS code.
- changed statecode default to "null" to match other inputs
- specified startDate and endDate if not populated in inputs (this is required if using readWQPdata()).
- Made readWQPsummary a flexible query in case site type or statecode are not included...saves if statements
- State code can be an input but not default.
- removed objects not needed in future computation (df_summary, sites; save space)
- removed temp RDS file creation/reading within package.
- added testthat back in that compares TADAdataRetrieval with TADABigdataRetrieval...does not depend upon statecode


cristinamullin (approved):

cristinamullin:
Example in documentation does not work (characteristic is required - can we update the error message?):

tada3 <- TADABigdataRetrieval(statecode = "CT")
Error in STUSAB %in% statecode : object 'STUSAB' not found
Called from: STUSAB %in% statecode

cristinamullin:
Instead of requiring characteristicname , can we make it more flexible, so that statecode, sitetype, or characteristicname (at least 1) could be provided?

ehinman:
@cristinamullin This function is ready for your review and testing. Appears robust on my machine but I am certain fresh eyes will identify weaknesses. Thank you!

ehinman:
It's a great question. I wonder what data submitters see in WQX when they upload a project ID. As a data user, I see the column names in the returned data profiles and it makes sense to me to have the function input match the column name on which I want to query. For example, I would be much more likely to know what a "ProjectIdentifier" input is, rather than "project". Similarly with "ActivityMediaName" rather than "sampleMedia". However, it is interesting that even the WQP browser interface is different. Continuing with these two examples, those fields are termed "Project ID" and "Sample Media" on the page itself. I think whichever we pick, adding documentation of field synonyms across querying platforms will make functions relatively simple to use (e.g. "project input is synonymous with the ProjectIdentifier column returned from the WQP and Project ID field on the web interface").

cristinamullin:
@ehinman I have another question related to this thread. Do you think it would make sense for TADAdataRetreival and TADABigdataRetrieval to be a single function? Where the chunking is only activated if needed, that is, if the summary service returns over X unique MonitoringLocationIdentifier's?

ehinman:
Interesting idea, and certainly can be done. The only downside I see is that running the summary will slow down the function, especially since we do not have the ability to specify the date range of interest in the summary service. Someone could specify they want one month of data for one characteristic name in one state and it'll pull data for that characteristic name and state over the entire time data are available. It's not likely to take too long, but it will add some time. Another consideration is the prevalence of sites with little to no data, such that they add to the site n count but may not require chunking to successfully return data. What do you think? Would you like me to try this? It would be so nice if there was a webservice that reported the size of a request so one (or a function!) could make a decision on which process to use.

PR #201, opened 2023-02-03
Update flag functions kh

katiehealy:
Updated flag functions so that when errorsonly = TRUE and there are no flagged rows in the dataframe, the result is an empty dataframe.


katiehealy:
@cristinamullin - I added options to the clean argument for the InvalidSpeciation function, but once again am running into the checkType issue (same as what happened with the InvalidCoordinates function). What was it that you did to resolve that issue?

cristinamullin (approved):

PR #205, opened 2023-02-09
Inputs spec chars eh

ehinman:
This branch consists of:
- Consistent date format YYYY-MM-DD in all functions and examples
- Column name inputs to names that match the query input (project and organization) for TADAdataRetrieval
- New special characters function that acts on a column input rather than hard coded ResultMeasureValue and DetectionQuantitationTypeLimit. MeasureValue 
- Spec char function also recognizes more special character situations (% and #,####) and flags them for the user
- Spec char function does not flag scientific notation as text
- Added lines to replace current special characters function to new function if desired/approved.


R/DataDiscoveryRetrieval.R
```diff
@@ -36,15 +36,15 @@
...
+#' @param startDate Start Date string in the format YYYY-MM-DD, for example, "2020-01-01"
 #' @param countycode Code that identifies a county 
 #' @param siteid Unique monitoring station identifier
 #' @param siteType Type of waterbody
 #' @param characteristicName Name of parameter
 #' @param sampleMedia Sampling substrate such as water, air, or sediment
-#' @param ProjectIdentifier A string of letters and/or numbers (some additional characters also possible) used to signify a project with data in the Water Quality Portal
-#' @param OrganizationIdentifier A string of letters and/or numbers (some additional characters also possible) used to signify an organization with data in the Water Quality Portal
-#' @param endDate End Date in the format YYYY-MM-DD
+#' @param project A string of letters and/or numbers (some additional characters also possible) used to signify a project with data in the Water Quality Portal
```
cristinamullin, line 45: Add note that the inputs for this are equivalent to the values that come out of the WQP profiles in the ProjectIdentifier field?

R/DataDiscoveryRetrieval.R
```diff
@@ -73,8 +76,8 @@ TADAdataRetrieval <- function(statecode = "null",
                               siteType = "null",
                               characteristicName = "null",
                               sampleMedia = "null",
-                              ProjectIdentifier = "null",
-                              OrganizationIdentifier = "null",
+                              project = "null",
+                              organization = "null",
```
cristinamullin, line 80: Same comment for OrganizationIdentifier (mention here in documentation & also in vignette?)

R/DataDiscoveryRetrieval.R
```diff
@@ -316,23 +319,22 @@ TADABigdataRetrieval <- function(startDate = "null",
...
+    startDate = lubridate::ymd("1800-01-01")
+    startYearLo = lubridate::year(startDate)
   } 
   
 # Logic: if the input endDate is not null, convert to date and obtain year
   # for summary
   if(!endDate=="null"){
-    endDate_High = lubridate::ymd(endDate)
-    endYearHi = lubridate::year(endDate_High)
+    endDate = lubridate::ymd(endDate)
```
cristinamullin, line 332: Should we mention anything Laura D shared here about the nuisances between NWIS, WQP services, and dataRetrieval in the documentation here as well (e.g. endDate and startDate are equal to endDate_High, startDate_Low). This may help some users, but also ourselves (to refresh our memories on why we made these decisions). This could also just be added as comments in the function itself instead of the documentation.

R/Utilities.R
```diff
@@ -64,15 +64,25 @@ autoclean <- function(.data) {
     toupper(.data$DetectionQuantitationLimitMeasure.MeasureUnitCode)
   # .data$BiologicalIntentName = toupper(.data$BiologicalIntentName)
   
-  # Remove duplicate rows
-  .data <- .data[!duplicated(.data), ]
+  # Remove duplicate rows - turned into a test because duplicated() takes a long
+  # time acting on all columns in a large dataset.
+  if(!length(unique(.data$ResultIdentifier))==dim(.data)[1]){
```
cristinamullin, line 69: Good thinking. This approach should work fine for removing exact duplicates. Let's add a test for this as well.

R/Utilities.R
```diff
@@ -64,15 +64,25 @@ autoclean <- function(.data) {
...
+    print("Duplicate records may be present. Filtering to unique records. This may take a while on large datasets.")
```
cristinamullin, line 70: change to "Duplicate records are present" ?

cristinamullin (changes requested):
Overall these changes look great! I added a few comments/suggestions for edits & will try to dig up some examples for you with exact duplicates and potential duplicates.

R/Utilities.R
ehinman: I wrote "may be present" because I could potentially see someone using autoclean on a dataset where they joined detection limit data to result data and thus have unique rows with the same result identifier and different detection limits. In this case, the function will check to make sure these cases are truly unique.

R/DataDiscoveryRetrieval.R
ehinman: I added additional commentary. Let me know if it is sufficient!

R/DataDiscoveryRetrieval.R
ehinman: See above!

R/DataDiscoveryRetrieval.R
ehinman: Added some details in the function description and in the vignette about date format. It is the one thing that isn't consistent between the WQP GUI and the package, but I think it's OK, given that the date format on the WQP GUI is a bit strange in the coding world. Let me know what you think.

R/Utilities.R
ehinman: I couldn't think of a good test for this so I went without. Let me know if you think of one.

cristinamullin (approved):

PR #234, opened 2023-03-17
Created TADA boxplot function

katiehealy:
TADA_boxplot() function takes dataset which has been harmonized and filtered to a specific ComparableDataIdentifier and creates a boxplot - much of the code was copied from CDC/ATSDR shiny app with a few modifications.

Note for future: create an .rda file that has been cleaned to run the example.


katiehealy:
@cristinamullin, how do I require importing the package "plotly"? is that done within the function roxygen skeleton or the NAMESPACE file?

cristinamullin:
> @cristinamullin, how do I require importing the package "plotly"? is that done within the function roxygen skeleton or the NAMESPACE file?

You can add this as a package dependency by including it in the description file under "Imports"

cristinamullin (approved):

PR #238, opened 2023-03-24
add hard coded limits

ehinman:
-added usgs limits not in wqx domain in the get ref function (Historical Lower Reporting Limit, Elevated Detection Limit). 
-updated ref csv as well.

### DIFF
--- a/R/GenerateRefTables.R
+++ b/R/GenerateRefTables.R
@@ -291,6 +291,13 @@ GetDetLimitRef <- function() {
       TRUE ~ as.character("Non-Detect")
     ))
   
+  ## Add USGS limits not in WQX domain table
+  usgs = data.frame(Name = c("Elevated Detection Limit","Historical Lower Reporting Limit"),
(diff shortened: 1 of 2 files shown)


cristinamullin:
Looks like we need to update the internal copy of the reference file (or may need to adjust this test since we are no longer using the domain table directly)?

── Failure ('test-GenerateRefTables.R:20:3'): Is GetDetLimitRef up to date? ────
old_latedate == new_latedate is not TRUE

ehinman:
@cristinamullin Whoops! Let me take a look.

ehinman (commented):
Hopefully this fixes the testing issue. We still use the domain table, the get ref function just adds some extra limits to the end of the domain table within the TADA package. The issue was that I didn't give the hard coded limits last change dates, which was throwing off the test. The test is built to make sure there aren't any new domain values from WQX that haven't been categorized as ND or OD, and it uses the "last change date" column to determine if the test passes or not.

cristinamullin (approved):

PR #239, opened 2023-03-24
Create WQXActivityTypeRef Table and Update QC Flag Function

katiehealy:
Created WQXActivityTypeRef table which categorizes Activity Codes as QC samples or NA. Also updated QualityControlActivity() function to use the Ref Table and changed clean to boolean.


R/GenerateRefTables.R
```diff
@@ -303,3 +303,118 @@ UpdateDetLimitRef <- function() {
...
+           "Quality Control-Calibration Check Buffer",
+           "Sample-Positive Control")
+  other <- c("Quality Control Sample-Other")
+  
+  WQXActivityTypeRef <- raw.data%>%
+    dplyr::mutate(TADA.ActivityType.Flag = dplyr::case_when(
+      Code %in% dup ~ "QC_duplicate",
+      Code %in% blank ~ "QC_blank",
+      Code %in% cal ~ "QC_calibration",
+      Code %in% other ~ "QC_other"
```
ehinman, line 407: To align with the other flag functions, suggest giving non-QC samples a Flag as well. Either "Not_QC" or similar.

ehinman (commented):
This looks great! I offered a couple of comments for the flagging column to be consistent with the other flagging functions. A few additional things:
- You could add a test for new domain values similar to the ones created for DetectionLimitTypeName and ResultDetectionCondition if you'd like.
- Add to vignette? 
- I think the coverage tests are failing because the package is missing documentation for TADA_Stats(), which can be fixed by running devtools::document().

R/GenerateRefTables.R
katiehealy: good call! - I thought about doing that but some other flag functions leave the flag column blank for valid samples - is this something we should change for all flag functions? or just the ones that use a reference table?

katiehealy:
The check issues resolved themselves with the latest commits (yay!) @ehinman, what file can I find the tests in for new domain values for DetectionLimitTypeName and ResultDetectionCondition?

ehinman:
Woop woop! You can find the test file here: tests/testthat/test-GenerateRefTables.R

PR #240, opened 2023-03-28
Vignette review

cristinamullin:


katiehealy:
FYI - I think the R-CMD checks are angry that I changed clean = "all" to clean = TRUE in the QualityControlActivity example. I get the error "match.arg(clean) : 'arg' must be NULL or a character vector" even though the function requires that clean be logical. Not sure if merging into develop will solve this issue like it has in the past - thoughts @cristinamullin?

cristinamullin:
> FYI - I think the R-CMD checks are angry that I changed clean = "all" to clean = TRUE in the QualityControlActivity example in the vignette. I get the error "match.arg(clean) : 'arg' must be NULL or a character vector" even though the function requires that clean be logical. Not sure if merging into develop will solve this issue like it has in the past - thoughts @cristinamullin?

I edited this line in the vignette for now to get around this issue (this tells the vignette to reference this branch directly): 

remotes::install_github("USEPA/TADA", 
                        ref = "VignetteReview",
                        dependencies = TRUE)

I also added a new test dataset with a lot of tribal data called "TribalNations".

Issue #57, opened 2022-04-15
HistogramAndBoxplot


cristinamullin:
Develop two functions to create a histogram and box plot: 
1) histogram
2) boxplot

Function inputs/required columns
- Group rows for plots by: TADA.ComparableDataIdentifier from Harmonization Template (https://github.com/USEPA/TADA/blob/develop/inst/extdata/HarmonizationTemplate.csv)
- Time Period
- Unit
- Result Values

Leverage NADA2 R package or other packages if possible:
https://cran.r-project.org/web/packages/NADA2/index.html

PR #261, opened 2023-04-28
Histogram function

katiehealy:
Created TADA_hist() function to create an interactive plotly histogram for filtered TADA dataframe. Addresses #57


inst/extdata/WQXDetectionQuantitationLimitTypeRef.csv
```diff
@@ -32,6 +33,7 @@
 "Detection/Quantitation Limit Type(DetectionQuantitationLimitTypeName)",17,"Statistical Uncertainty","That component of uncertainty which arises from imprecision. www.iaea.org/OurWork/ST/NE/NEFW/WTS-Networks/IDN/idnfiles/Wkp_on_Development_WM,WasteCharacterization_Clearance/Day3/12.Statistics_GREY.pdf","12/9/2015 7:55:38 AM","Other"
 "Detection/Quantitation Limit Type(DetectionQuantitationLimitTypeName)",18,"Systematic Uncertainty","That component of uncertainty which arises from biases. www.iaea.org/OurWork/ST/NE/NEFW/WTS-Networks/IDN/idnfiles/Wkp_on_Development_WM,WasteCharacterization_Clearance/Day3/12.Statistics_GREY.pdf","12/9/2015 7:55:38 AM","Other"
 "Detection/Quantitation Limit Type(DetectionQuantitationLimitTypeName)",28,"Taxonomic Loss Threshold","Taxonomic Loss Threshold","2/16/2023 12:05:38 PM","Other"
+"Detection/Quantitation Limit Type(DetectionQuantitationLimitTypeName)",37,"Upper 95% Confidence Limit","Upper limit of 95% confidence limit, the above point estimate within which the true value ","4/25/2023 2:58:03 PM","Non-Detect"
```
cristinamullin, line 36: Should this be referenced as an over-detect (above upper limit)?

cristinamullin:
Nice work.! I left one comment with a question about one of the detection limit assignments. Can you please replace the histogram in the vignette with this new function too? Thanks!

inst/extdata/WQXDetectionQuantitationLimitTypeRef.csv
katiehealy: Good catch! - I will edit the GetDetLimitRef function

cristinamullin: After reviewing this again, I'm actually not sure if these detection limit values (95% confidence limit) should be assigned to the above or below detection groups. The definition looks incomplete, so it is not clear to me if these values correspond to the device/sample detection limits or not. Did you see these values used in an example dataset? If yes, could you share the dataset? @ehinman did you leave these out on purpose initially? Any thoughts?

katiehealy: I don't think this issue existed before - when I was running through the checks for the histogram function, the tests failed bc the Detection Limit Reference Table needed to be updated (I think maybe it was updated by the WQX team recently? not sure..) so I ran the UpdateDetLimitRef function which added the new item for upper 95% confidence limit

katiehealy: The domain table description is rather confusing.. "the above point estimate within which the true value" - is it missing some words at the end?

cristinamullin: Yes, nice to see the detection limit reference table test is working well! I agree both the upper and lower 95% fit best in the "Other" category

cristinamullin: Once you make this change, don't forget to run pkgdown::build_site() to update TADA GitHub pages site. After that, we should be ready to merge. Thanks again Katie!

ehinman:
@katiehealy Are you still working on this or looking for a review?

katiehealy:
@ehinman I have some more commits, but the vignette keeps failing to build because the TADA_hist() function doesn't exist in develop yet. I might just comment it out for now until it's fully merged into develop? check() is also finding some errors in the tests for the censored data suite - not sure what's going on there.

ehinman:
@katiehealy Ok! Sounds good. Hmm...not sure on the censored data test error messages... One way I got around the vignette build error is that I would temporarily change the reference repo from develop to my working branch like remotes::install_github("USEPA/TADA", ref = "elise_branch")). That then has to be changed back to develop after it is merged in.

R/GenerateRefTables.R
```diff
@@ -287,7 +287,7 @@ GetDetLimitRef <- function() {
   WQXDetLimitRef <- raw.data%>%
     dplyr::mutate(TADA.Limit_Type = dplyr::case_when(
       Name%in%c("Upper Quantitation Limit","Upper Reporting Limit","Upper Calibration Limit") ~ as.character("Over-Detect"),
-      Name%in%c("Drinking Water Maximum","Field Holding Time Limit","Specified in workplan","Statistical Uncertainty","Systematic Uncertainty","Taxonomic Loss Threshold","Water Quality Standard or Criteria") ~ as.character("Other"),
+      Name%in%c("Drinking Water Maximum","Field Holding Time Limit","Specified in workplan","Statistical Uncertainty","Systematic Uncertainty","Taxonomic Loss Threshold","Water Quality Standard or Criteria","Upper 95% Confidence Limit","Lower 95% Confidence Limit") ~ as.character("Other"),
```
ehinman, line 290: Can you also add the "Upper 95% Confidence Limit" to the other category in this function, and then run UpdateDetLimitRef() to update the .csv?

vignettes/TADAModule1.Rmd
```diff
@@ -213,7 +224,7 @@ to dates, datetimes, and numerics based on a standard algorithm.
 Enter ?TADAdataRetrieval into the console to review more example queries
 and additional information.
 
-```{r}
+```{r TADAdataRetrieval}
 TADAProfile <- TADAdataRetrieval(organization = c("REDLAKE_WQX", "SFNOES_WQX", "PUEBLO_POJOAQUE", "FONDULAC_WQX", "PUEBLOOFTESUQUE", "CNENVSER"), startDate = "2018-01-01")
```
ehinman, line 228: This is a larger dataset example, which is really nice for trouble shooting and scoping what a user might experience. However, it takes forever to run through the vignette with all of the steps and devtools::check() takes 30+ minutes. Might suggest cutting it down to 2 years of data, or less.

R/Figures.R
```diff
@@ -82,13 +82,104 @@ TADA_boxplot <- function(filtered.data) {
...
+  if (length(unique(filtered.data$TADA.ComparableDataIdentifier)) > 1) {
+    warning("Histogram function cannot run with more than 1 comparable data identifier. Please filter dataframe and rerun function.")
+  }
+  
+  # execute function after checks have passed
+  
+  # units
+  unit <- unique(filtered.data$TADA.ResultMeasure.MeasureUnitCode)
+  char <- unique(filtered.data$TADA.CharacteristicName)
+  x_label <- paste0("Level of ", char, " (", unit, ")")
```
ehinman, line 132: Suggest getting rid of "Level of ", since the x-axis is (usually) the concentration of the characteristic. I think you could just show char (unit).

R/GenerateRefTables.R
katiehealy: I think "Upper 95% Confidence Limit" is now in the "other" category, but I'll rerun UpdateDetLimitRef()!

vignettes/TADAModule1.Rmd
katiehealy: Good idea - it does take forever. I'll put an end date of January 1, 2020, and see if that helps.

R/Figures.R
```diff
@@ -82,13 +82,104 @@ TADA_boxplot <- function(filtered.data) {
...
+      hoverlabel=list(bgcolor="white"),
+      title = paste0("Boxplot of ", char, " (", unit, ")"), 
+      plot_bgcolor = "#e5ecf6"
     ) %>% 
     plotly::config(displayModeBar = FALSE)
   
   return(base_boxplot)
 }
 
+#' Create Histogram
```
ehinman, line 94: These plots look really great!! My only suggestion would be to remove the dependency of the histogram and boxplot on the TADA.ComparableDataIdentifier, and instead create another input for this plot that is something along the lines of id_col = comparables column OR characteristic column. I could see a user wanting to visualize data prior to running the harmonization step, so it would be nice to give them the option to just filter to a specific characteristic name. You could make the documentation say that id_col can equal either TADA.DataComparableIdentifier OR TADA.CharacteristicName, and then it can make sure that the one the user picked is in the dataset. Currently, it is also problematic that it requires one TADA.ComparableIdentifier but then uses the TADA.CharcteristicName, which may not be unique to the dataset (since the comparable identifier allows users to group multiple characteristic names into one group).

ehinman (commented):
This looks really good, and you're so right, you have both Upper and Lower 95% Confidence Limit already in the other category.

R/Figures.R
katiehealy: ooh good point! hmm, I had originally created the boxplot and histogram functions under the assumption that it would be required that the datasets be harmonized in case units of measure were not consistent across one characteristic. Do you think that is not usually an issue across one characteristic? Or should I just create an error/warning if there is more than one unit of measure?

ehinman: Oooh also a good point. Shoot. Yeah, I would probably just create a warning if there are multiple units (in any situation, but I guess would only happen if id_col = TADA.CharName), multiple characteristic names (if id_col = TADA.CharacteristicName), or multiple comparable data identifiers (if id_col = "TADA.ComparableDataIdentifier"), and then make sure the plot labels reflect the id_col input. I think the vision is to leverage these in TADAShiny, too, and in that interface, we can make users pick a unique characteristic + unit/comparable combination to be fed into the plotting functions. I'm not exactly sure the full implementation of the harmonization template in the QC module, and how much of it will be user-populated versus TADA-suggested (currently, we have some nutrient-related suggestions with everything else having a TADA.ComparableDataIdentifier of NA and I recall Cristina wanted to do some more work with it), but I think making the plots more flexible to be used throughout the process would be a nice addition. What do you think?

katiehealy: that sounds like a good idea! I'll add the input column options and unit error message :)

ehinman: Sounds good, thanks!

ehinman:
@katiehealy This all looks good. I made some small changes to the plotting documentation and changed some error/warning messages. Do you want to merge this pull request to pseudo-develop?

ehinman:
Sounds good! Feel free ready you're ready.

PR #262, opened 2023-05-05
utilities updates

ehinman:
- ConvertSpecChararacters now alerts users when they have non-ASCII characters in their result measure values (cannot be properly handled by R)
- Added idCensoredData to autoclean
- Small fix on one of the flagging functions
- Made OrderTADAColumns also order rows by ResultIdentifier. Required a test update as well.

I do not think this will currently pass the tests, since I started to also change the update detection limit ref files and realized @katiehealy was covering that in her pull request. We can test pseudo-develop after we've merged our changes.


cristinamullin:
Regarding this first one: ConvertSpecChararacters now alerts users when they have non-ASCII characters in their result measure values (cannot be properly handled by R)

Elise, did you already mention this to Adam and Kevin (WQX and WQP teams)? We may be able to control/fix what gets into the WQP through WQX and on the USGS side as well. Ideally, these should not be able to go through to WQP as non-ASCII characters (should be flagged/fixed on data submission side).

PR #264, opened 2023-05-08
update convert result units

ehinman:
when transform = TRUE, only identifies results where conversion factor was not NA or 1: 1 indicates no conversion needed.


ehinman:
I think this is small enough, I am just going to merge it in!

PR #266, opened 2023-05-10
Update Utilities.R

ehinman:
I noticed that some of the functions re-arrange original columns in kind of strange ways (due to flagging joins). This commit ensures the data retrieval columns retain their order in addition to the tada columns. Any weird columns not contained in these lists will end up in the middle.


katiehealy (approved):
looks good to me! (just will have to resolve the conflicts since you fixed the bug in the other branch)

PR #267, opened 2023-05-12
Char ranges eh

ehinman:
Can you take a look and let me know if anything looks wonky? I ran the tests and devtools::check() and everything seems fine, but would like another set of eyes!

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -37,6 +37,7 @@ Imports:
     plyr, 
     data.table, 
     dplyr,
+    tidyr,
     grDevices,
     magrittr,
     stringr,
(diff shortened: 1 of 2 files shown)


katiehealy:
It looks good to me, but I tried it on a couple example data sets and I'm running into this error with OrderTADACols at the end:
Error in `dplyr::arrange()`:
! Can't transform a data frame with duplicate names.

ehinman:
@katiehealy Thanks for finding this bug! Apparently, it's potentially an issue with the rlang package: https://community.rstudio.com/t/cant-transform-a-data-frame-with-duplicate-names/131688

I switched to a base R command instead, since I worry others might run into the same problem.

katiehealy:
example dataset testing worked! I think you're good to merge :)

katiehealy (approved):

ehinman:
Thank you!

PR #268, opened 2023-05-15
Filter alt eh

ehinman:
@katiehealy I added some filter functions that I think have almost identical functionality to what is already in the package, but are much faster/simpler to run. Would be interested to hear what you think.


R/Filtering.R
katiehealy, line 1: looks good! these will be super useful :) a couple thoughts:
- for filterCounts() display argument, you could put all the options in the argument and use match.arg to issue a more clear warning if someone tries to enter an invalid option. so in the arguments, it would be display = c("key", "narrow", "all") instead of display = "key". "key" will still be the default option because it's first in the list. and then in the checks section you would add a line: display <- match.arg(display)
- for some reason, when I intentially misspell a characteristicName in either fieldCounts or filterPie it's not showing me the warning you created: "Characteristic name(s) provided are not contained within the input dataset. Note that TADA converts characteristic names to ALL CAPS for easier harmonization." instead, it's showing either a dataset of zero rows or "Error in names(df) %in% cols : object 'cols' not found"

ehinman:
Thanks for the review! I think I caught the issue with characteristicName and I like the match.arg suggestion, so I added that too.

katiehealy (approved):
error message works now!

PR #283, opened 2023-06-06
id nearby sites and id potential duplicates

ehinman:
added two functions: one that identifies nearby sites and puts them into groups, and one that identifies potentially duplicated results based on date, time, characteristic, unit, and site locations. This catches when multiple organizations upload the same datasets. Example: state of PA and USGS-PA water quality sampling.


R/ResultFlagsIndependent.R
```diff
@@ -1131,30 +1131,80 @@ InvalidCoordinates <- function(.data,
...
-  dcoords = sf::st_as_sf(x = distances, coords = c("LongitudeMeasure","LatitudeMeasure"), crs="EPSG:4326")
-  
-  dists = data.frame()
-  for(i in 1:max(dcoords$dup_id)){
-    ds = subset(dcoords, dcoords$dup_id==i)
-    dist = as.numeric(sf::st_distance(ds$geometry[1],ds$geometry[2]))
-    dsdist = data.frame(dup_id = i, distance_m = as.numeric(dist))
-    dsdist$TADA.idPotentialDuplicates.Flag = ifelse(dist<=dist_buffer,"POTENTIAL DUPLICATE DATAPOINT",NA)
-    dists = rbind(dists, dsdist)
+idPotentialDuplicates <- function(.data, dist_buffer = 100){
```
cristinamullin: Would it be difficult to add a function input here to look: 1) only within an organization for potential duplicates, or 2) across all organizations? Alternatively, not restrict this functionality to only look across organizations, but also within an org? There are scenarios where the same organization will upload the same historical data with a different site name/slightly different lat/longs as well (legacy vs current sites)

ehinman: I'd have to think through this a little, but probably. They could use TADA_NearbySites to find historic/current monitoring locations and see if duplicated data submitted across those. I noticed that exact situation in a test I did yesterday with TribalData.

ehinman: Are you thinking that the within org test would loop through each org and check all the other columns + nearby sites? That wouldn't be too bad, I don't think.

R/Utilities.R
```diff
@@ -646,3 +646,93 @@ getTADATemplate <- function(){
...
+
+
+#' Identify and group nearby monitoring locations
+#'
+#' This function takes a TADA dataset and creates a distance matrix for all sites in the dataset. It then uses the buffer input value to determine which sites are within the buffer and should be identified as part of a nearby site group.
+#'
+#' @param .data TADA dataframe OR TADA sites dataframe
+#' @param dist_buffer Numeric. The maximum distance (in meters) two sites can be from one another to be considered "nearby" and grouped together.
+#'
+#' @return Input dataframe with one or more TADA.SiteGroup columns (depending upon if a site belongs to more than one nearby group), which identify each site group as "Group_#" for easy filtering.
```
cristinamullin, line 570: discuss automatically updating TADA.MonitoringLocationIdentifier to the new MonitoringLocationIdentifier for the grouping, if user desires to assess the grouping as a "single" site

R/ResultFlagsIndependent.R
cristinamullin: That's good you found an example dataset with this issue for testing/dev. I think SC may be another good test state for this issue. That approach sounds like it would work

cristinamullin (approved):

PR #288, opened 2023-06-13
Fixing minor error in InvalidResultUnit

katiehealy:
InvalidResultUnit wasn't retaining the TADA.ResultUnit.Flag column for dataframes where there were no flagged results - this fixes that issue.


ehinman (approved):
This looks great @katiehealy! Thanks so much for catching this.

PR #303, opened 2023-07-06
auto styler

cristinamullin:
https://styler.r-lib.org/reference/style_pkg.html

ran with defaults:
> library(styler)
> style_pkg()

### DIFF
--- a/R/Tables.R
+++ b/R/Tables.R
@@ -1,150 +1,151 @@
 #' @title Summarize data downloaded by specified column
-#' 
-#' @description 
-#' Function to summarize the number of sites and records downloaded from the 
+#'
+#' @description
+#' Function to summarize the number of sites and records downloaded from the
 #' WQP for each unique column group.
-#' 
+#'
 #' @param .data TADA data frame containing the data downloaded from the WQP, where
 #' each row represents a unique data record.
 #' @param col A text string name of the column the user would like summarized.
-#' 
+#'
 #' @return A dataframe containing a column for each unique element, the number of
 #' sites with that element populated, and the number of records with that element
 #' populated.
-#'  
+#'
 #' @export
-#'  
+#'
 
-TADA_SummarizeColumn <- function(.data,col="TADA.CharacteristicName"){
-  .data$summ = .data[,col]
+TADA_SummarizeColumn <- function(.data, col = "TADA.CharacteristicName") {
+  .data$summ <- .data[, col]
   # Summarize WQP data pull
   wqp_summary <- .data %>%
     dplyr::group_by(summ) %>%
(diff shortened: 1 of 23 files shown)


cristinamullin:
@jbousquin @ehinman what do you think?

ehinman:
I think anything that adds consistency to our package and makes it more easily comprehended is a worthwhile addition. This styling looks fine to me!

R/Tables.R
```diff
@@ -1,151 +1,152 @@
...
-  
-  sum_hi = cens%>%dplyr::group_by_at(spec_cols)%>%
-    dplyr::filter(TADA.CensoredData.Flag%in%c("Over-Detect", "Uncensored"))%>%
-    dplyr::summarise(Measurement_Count = length(unique(ResultIdentifier)), Censored_Count = length(TADA.CensoredData.Flag[TADA.CensoredData.Flag=="Over-Detect"]), Percent_Censored = length(TADA.CensoredData.Flag[TADA.CensoredData.Flag=="Over-Detect"])/length(TADA.CensoredData.Flag)*100, Censoring_Levels = length(unique(TADA.ResultMeasureValue[TADA.CensoredData.Flag=="Over-Detect"])))%>%
-    dplyr::filter(Censored_Count>0)%>%
+
+  sum_hi <- cens %>%
+    dplyr::group_by_at(spec_cols) %>%
+    dplyr::filter(TADA.CensoredData.Flag %in% c("Over-Detect", "Uncensored")) %>%
+    dplyr::summarise(Measurement_Count = length(unique(ResultIdentifier)), Censored_Count = length(TADA.CensoredData.Flag[TADA.CensoredData.Flag == "Over-Detect"]), Percent_Censored = length(TADA.CensoredData.Flag[TADA.CensoredData.Flag == "Over-Detect"]) / length(TADA.CensoredData.Flag) * 100, Censoring_Levels = length(unique(TADA.ResultMeasureValue[TADA.CensoredData.Flag == "Over-Detect"]))) %>%
```
jbousquin, line 135: This line is long, it was before too, just tells us the styler isn't enforcing line char limit

jbousquin:
Agreed looks like a great way to increase consistency. Only concern would be if a style change broke something not currently covered by tests. So far all the changes I've looked at were '<-' or white space. I can keep looking through as part of review, anything additional to draw attention to or files that don't have tests/ are known to be finicky?

R/Tables.R
ehinman: Oooh good point--something to look into. It looks like this comparison view is trying to ensure each line is the same width on the screen...I think this is indicating that styler added a line after cens%>% on line 132, which moves everything else down. There appears to be a big gap on line 134 on the "after" pane because the "before" code made line 134 longer, and in the "after" it's all on line 135.

PR #305, opened 2023-07-06
Exports and data names

cristinamullin:
Removed cached data function roxygen so they will not show up as exported package functions. Updated example data names.


cristinamullin:
@ehinman can you help make Shiny match these new example data file names?

ehinman:
@cristinamullin Sure!

R/ExampleData.R
```diff
@@ -1,89 +1,89 @@
...
-#' TADAProfileCleanTP
+#' Data_TP_6Tribes_5y
 #' 
-#' A dataset containing Total Phosphorus data from Tribal Data (see ?TribalData for more information):
-#'
-#'
-#' Data was cleaned using the TADA harmonize vignette. Data is filtered down to 
-#' only Total Phosphorus, Mixed Forms.
+#' A dataset containing Total Phosphorus (TP) data from tribal organizations
+#' (see ?TribalData for more information). Data was cleaned using the
```
ehinman, line 71: I think this needs to be updated to the new name Data_TP_6Tribes_5y.

ehinman (commented):
It looks like the datasets (.Rda) themselves still need to be renamed. I think that will resolve the errors in the tests.

PR #310, opened 2023-07-13
Within org dups eh

ehinman:
- created new function that identifies duplicates within organizations at the same site, date, time, characteristic, fraction, and depth. Should be used to identify multiple uploads of the same results. 
- Renames ID duplicate functions to discern them
- Should they be combined? 
- Also created a new function for creating random datasets to test against TADA functions


R/ResultFlagsIndependent.R
```diff
@@ -1021,48 +1022,117 @@ TADA_FindPotentialDuplicates <- function(.data, dist_buffer = 100){
...
+#'   TADA.ResultSelectedWithinOrg, which defaults to TRUE (all results
+#'   selected), unless the user specifies the handling method as 'pick_one', in
+#'   which case there exists one result in each duplicate group where
+#'   TADA.ResultSelectedWithinOrg = TRUE and for all other results in the group
+#'   TADA.ResultSelectedWithinOrg = FALSE.
+#' 
+#' @export
+#' 
+
+TADA_FindPotentialDuplicatesWithinOrg <- function(.data, handling_method = 'none'){
```
cristinamullin, line 1099: If combined: 

handling_method could be randomly select 1, choose 1 monitoringlocationID of the two (single org dup), or choose 1 organizationID of the two (multiple org dup).

The distance buffer would be helpful for both of these. Within a single org, there may be a retired and current site that may not be at exactly the same point, but have duplicates of the historical data.

R/ResultFlagsIndependent.R
```diff
@@ -978,28 +978,29 @@ TADA_FlagCoordinates <- function(.data,
 #' be flagged as potential duplicates.
 #'
 #' @return The same input TADA dataframe with additional columns: a
-#'   TADA.ProbableDuplicate column indicating if there is evidence that two or
-#'   more results are duplicated, a TADA.DuplicateID column, containing a
+#'   TADA.ProbableBetweenOrgDuplicate column indicating if there is evidence that two or
+#'   more results are duplicated, a TADA.BetweenOrgDupGroupID column, containing a
```
cristinamullin, line 982: suggestion:

"TADA.MultipleOrgDuplicate column indicating if there is evidence that results are likely duplicated due to submission of the same dataset by two or more different organizations, a TADA.MultipleOrgDupGroupID column"

R/ResultFlagsIndependent.R
```diff
@@ -964,7 +964,7 @@ TADA_FlagCoordinates <- function(.data,
   return(.data)
 }
 
-#' Identify Potential Duplicate Data Uploads (UNDER ACTIVE DEVELOPMENT)
+#' Identify Potential Duplicate Data Uploads Between Organizations (UNDER ACTIVE DEVELOPMENT)
```
cristinamullin, line 967: Identify Potential Duplicate Data Uploads by Multiple Organizations

R/Utilities.R
```diff
@@ -720,3 +722,25 @@ TADA_FindNearbySites <- function(.data, dist_buffer=100){
   
   return(.data)
 }
+
+# Generate a random data retrieval dataset (internal, for testthat's)
+# samples a random state and a random 3 months in the past 10 years using
+# TADA_DataRetrieval. Built to use in testthat and for developer testing of new
+# functions on random datasets.
+
+TADA_RandomTestingSet <- function(number_of_days = 90){
```
cristinamullin, line 731: This is great!

cristinamullin:
These could potentially be combined, but the handling methods would differ. For that reason, it might make more sense to keep them separate for now. 

Is it possible to add a handling method for within an org: IF the duplicates are all from two different sites, randomly select or have user select to use all from one of the two monitoringlocationIDs .

Across orgs, could the user select to use all data from one of the two organizations, or randomly select one?

R/ResultFlagsIndependent.R
ehinman: We can discuss further, but my initial reaction is that the first suggestion would be difficult to execute. I'm not sure how someone would pick a single monitoring location or organization when every single duplicate group has potentially unique comparisons (i.e. duplicate groups comparing between orgs will not necessarily always have the same two or more organizations within the duplicate group; every duplicate group has a set of unique monitoring locations....not sure how you'd confidently pick one naming convention that applies to all sites in each duplicate group.)

Re: second suggestion, I think there is a low likelihood that a data uploader would upload the same data result twice for a current and a retired monitoring location. I'm sure it has happened, but I wasn't sure it was worth creating a function for it. Do you think the need is great enough to craft this additional function? It might be easier to just use the TADA_IDNearbySites function and custom group them in the second module.

R/ResultFlagsIndependent.R
```diff
@@ -1021,48 +1022,117 @@ TADA_FindPotentialDuplicates <- function(.data, dist_buffer = 100){
...
+#' 
+#' @param .data TADA dataframe
+#' @param handling_method Defaults to 'none'. When input is 'none', function
+#'   flags groups of potentially duplicated results, but makes no decisions on
+#'   which (if any) should be selected/excluded. Alternatively, user may
+#'   populate handling_method = 'pick_one' and function will randomly select one
+#'   result in each group and flag all others as duplicates.
+#' 
+#' @return The same input TADA dataframe with additional columns: a
+#'   TADA.WithinOrgDupGroupID column indicating whether a result is part of a
```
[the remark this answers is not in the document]
ehinman: Hmm...I don't know which I like better. Single and Multiple works okay, though.

ehinman:
See message above. I think your suggestion for selecting an org from the multiple org dup function could easily be executed in the shiny app, but I struggle to figure out a great way to implement this as a function in TADA, since the user would need to know their priority org before running the function, and use their name as an input. Then, there would need to be a rule if a duplicate group didn't have the priority org....then pick randomly? Definitely do-able, but would be a fairly technical function that would need some additional work on the user's part.

ehinman:
Hm, with more consideration, the TADA function could:

1. allow the user to input an organization hierarchy (e.g. org_hierarchy = c("USGS-PA", "PADEP", "CityOfPhilly"...etc.)) OR
2. have the function pick a result randomly OR
3. keep all results (do not flag duplicates, only identify them)

cristinamullin (approved):

PR #311, opened 2023-07-13
update figures

ehinman:
- figures require comparable data identifier but can also be grouped by additional columns. the output is a list of figures for all unique groups in the dataset.


R/Figures.R
katiehealy, line 1: These look great! I love the expanded functionality. My only suggestion would be to include an example of using multiple id_cols (maybe MonitoringLocationTypeName?)

katiehealy (approved):

R/Figures.R
cristinamullin: Good idea!

ehinman: I agree. I will add some!

cristinamullin (approved):

PR #330, opened 2023-08-18
Smol updates

ehinman:
- updates following NP discussion, plus some small figure changes


ehinman:
I realized that if the equation accepts any combination of subsets of subspecies, other equations later in the hierarchy might be completely skipped even though a more complete set of subspecies exist to evaluate a later equation. To fix this, I had to basically give the function every possible combination of nutrient subspecies summations, in a larger hierarchy, and have each loop look for situations where ALL subspecies are present. This is now saved as a .csv in the package (NP_equations.csv).

For example, if a nutrient grouping on a single day/site had NITRATE and AMMONIA, because the TKN + NITRATE + NITRITE equation comes earlier in the hierarchy, the function would accept ONLY NITRATE to represent TOTAL N, when really it should skip down to the AMMONIA + NITRATE + NITRITE equation and work with those subspecies together. This is now fixed, and it will sum AMMONIA and NITRATE, as intended.

cristinamullin (approved):

PR #331, opened 2023-08-23
Pair Replicates Function

katiehealy:
Created function to pair replicates with their parent samples. (This function only pairs, does not calculate RPD of measured values.)


R/ResultFlagsDependent.R
```diff
@@ -548,6 +548,126 @@ TADA_FindQCActivities <- function(.data, clean = FALSE, flaggedonly = FALSE) {
...
+#' "Quality Control Field Replicate Habitat Assessment",
+#' "Quality Control Field Replicate Msr/Obs",
+#' "Quality Control Field Replicate Portable Data Logger",
+#' "Quality Control Field Replicate Sample-Composite", and
+#' "Quality Control Sample-Field Replicate".
+#' @param time_window Numeric argument defining the time window in seconds within 
+#' which to search for parent samples. The default time window is 300 seconds or 5 minutes.
+#' 
+#' @return This function adds two columns to the original data frame: 'TADA.DateTime' and 
+#' 'TADA.ReplicateSampleID'. 'TADA.DateTime' combines the columns 'ActivityStartDate' and
```
cristinamullin, line 572: Creating TADA.DateTime may not be necessary. See: ActivityStartDateTime. Does that give you the date time in the format you need?

R/ResultFlagsDependent.R
```diff
@@ -548,6 +548,126 @@ TADA_FindQCActivities <- function(.data, clean = FALSE, flaggedonly = FALSE) {
...
+#' # Run TADA_FindQCActivities:
+#' df <- TADA_FindQCActivities(Data_NCTCShepherdstown_HUC12)
+#' 
+#' # Find pairs for all data flagged as "QC_replicate":
+#' df_all_pairs <- TADA_PairReplicates(df)
+#' 
+#' # Find pairs for only data with ActivityTypeCode "Quality Control Sample-Field Replicate":
+#' df_fieldrep_pairs <- TADA_PairReplicates(df, type = "Quality Control Sample-Field Replicate")
+#' 
+#' # Find pairs for all data flagged as "QC_replicate" with a 10-minute time window:
```
cristinamullin, line 598: Suggestion to clarify in documentation: By 10-min time window, does that mean 5 mins before or after the paired sample (10 min window), or 10 mins before or 10 mins after (20 min window)?

R/ResultFlagsDependent.R
katiehealy: Oh, good call! I didn't notice that column before. Let me try it out and see what happens.

R/ResultFlagsDependent.R
katiehealy: Good idea - right now, if time_window = 600 (10 minutes), that means it will find potential pairs 10 minutes before or 10 minutes after, so a 20 minute window. Any thoughts on a better word? maybe time_difference?

cristinamullin: time_difference makes sense to me. It may be helpful to say in the documentation that this is the max time difference between the paired sample and replicate. From that, the time window that the function looks at for each sample can be inferred (e.g. if the max time difference is 5 mins, then the time window that the function searches for is 10 mins (i.e., 5 mins before and 5 mins after the parent sample time).

cristinamullin (approved):

PR #369, opened 2023-12-20
358 measurequalifiercode definitions

hillarymarler:
For TADA_MeasureQualifierCodeFlag adds an argument "define" with default set to to TRUE which creates a column called "TADA.MeasureQualifierCode.Define" which concatenates the qualifier code with its definition. This accommodates more than one qualifier code and will return definitions for any codes which have a matching description in the reference file.

I also updated TADA.MeasureQualifierCode.Flag to consider each qualifier code individually (rather than relying on hard-coded combinations). This required creating a hierarchy to assign flags if multiple codes had conflicting flags for one result. Current hierarchy is Suspect > Non-Detect > Over-Detect > Pass > Not Reviewed.

I also updated the search for missing codes to search by single codes, not combinations.

### DIFF
--- a/R/ResultFlagsDependent.R
+++ b/R/ResultFlagsDependent.R
@@ -626,7 +626,7 @@ TADA_AutoFilter <- function(.data) {
 #' rows of data flagged as Suspect.
 #' 
 #' @param define Boolean argument; the default is define = TRUE. When define = TRUE,
-#' the function will add an additional column (TADA.MethodQualifierCode) providing 
+#' the function will add an additional column (TADA.MethodQualifierCode.Def) providing 
 #' all available definitions for the MethodQualifierCodes for each result. When 
 #' define = FALSE, no additional column is added.
 #'
@@ -690,9 +690,9 @@ TADA_FlagMeasureQualifierCode <- function(.data, clean = FALSE, flaggedonly = FA
      tidyr::unnest(MeasureQualifierCode) %>%
      merge(mqc.ref) %>%
      dplyr::group_by(ResultIdentifier) %>%
-     dplyr::summarize(TADA.MeasureQualifierCode.Define = paste(Concat, collapse = "; "))
+     dplyr::summarize(TADA.MeasureQualifierCode.Def = paste(Concat, collapse = "; "))
 
-   .data$TADA.MeasureQualifierCode.Define <- mqc.TADA$TADA.MeasureQualifierCode.Define[match(.data$ResultIdentifier, mqc.TADA$ResultIdentifier)]
+   .data$TADA.MeasureQualifierCode.Def <- mqc.TADA$TADA.MeasureQualifierCode.Def[match(.data$ResultIdentifier, mqc.TADA$ResultIdentifier)]
 
    rm(mqc.ref, mqc.TADA)
  }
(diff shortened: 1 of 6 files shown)


cristinamullin:
This all sounds great – thanks Hillary! The hard coded combinations in the WQXMeasureQualifierCodeRef ref are no longer needed, correct? You can remove them by removing lines 735-783 from WQXRefTables.R in the TADA_GetMeasureQualifierCodeRef function. Once those lines are removed, you can then run TADA_UpdateMeasureQualifierCodeRef() to update the reference file.

cristinamullin:
@hillarymarler this should fix the two warnings: 

1) add purrr under Imports in the Description file
2) update docs (see below)
`
library(devtools)
`
`
load_all()
`
`
document()
`

❯ checking dependencies in R code ... WARNING
  '::' or ':::' import not declared from: 'purrr'
  Namespace in Imports field not imported from: 'remotes'
    All declared Imports should be used.

❯ checking for code/documentation mismatches ... WARNING
  Codoc mismatches from documentation object 'TADA_FlagMeasureQualifierCode':
  TADA_FlagMeasureQualifierCode
    Code: function(.data, clean = FALSE, flaggedonly = FALSE, define =
                   TRUE)
    Docs: function(.data, clean = FALSE, flaggedonly = FALSE)
    Argument names in code not in docs:
      define

hillarymarler:
Thanks - I will update to fix the warnings and edit/update the reference table.

R/ResultFlagsDependent.R
```diff
@@ -624,6 +624,11 @@ TADA_AutoFilter <- function(.data) {
 #' @param flaggedonly Boolean argument; the default is flaggedonly = FALSE. When
 #' flaggedonly = TRUE, the function will filter the dataframe to show only the
 #' rows of data flagged as Suspect.
+#' 
+#' @param define Boolean argument; the default is define = TRUE. When define = TRUE,
+#' the function will add an additional column (TADA.MethodQualifierCode) providing
```
cristinamullin, line 629: Suggest to call the new field:
TADA.MeasureQualifierCode.Def

hillarymarler: Sounds good to me - I will make that change now.

cristinamullin (approved):

R/ResultFlagsDependent.R
```diff
@@ -665,26 +670,71 @@ TADA_FlagMeasureQualifierCode <- function(.data, clean = FALSE, flaggedonly = FA
...
+  codes <- stringr::str_split(unique(.data$MeasureQualifierCode), ";") %>%
+    unlist() %>%
+    unique()
+  
   if (any(!codes %in% qc.ref$MeasureQualifierCode)) {
     missing_codes <- codes[!codes %in% qc.ref$MeasureQualifierCode]
     missing_codes_df <- data.frame(
       MeasureQualifierCode = missing_codes,
-      TADA.MeasureQualifierCode.Flag = "uncategorized"
+      TADA.MeasureQualifierCode.Flag = "uncategorized",
```
cristinamullin, line 730: I noticed in other flag functions that depend on reference tables we are using the term "Not Reviewed" which is similar to "uncategorized". Should we be consistent across all reference files and in what is delivered in all TADA flag fields?

PR #370, opened 2023-12-21
323 identifying censored data

hillarymarler:
I used TADA_FlagMeasureQualifierCode to identify additional results that should be considered in the TADA_IDCensoredData function. These additional results are subject to the same logic tests regarding detection types and limit types as results identified using TADA.ResultMeasureValueDataTypes.Flag == "Result Value/Unit Copied from Detection Limit". 

This means that some of the additional results now included in TADA_IDCensoredData end up with the TADA.CensoredData.Flag: "Detection condition is missing and required for censored data ID." or "Conflict between Condition and Limit.". Some of these may be resolved in the future as we are able to consider more than detection limit in the future. For now, I have left this as is and the logic behind assigning the TADA.CensoredDataFlag remains the same whether the results are identified via TADA.ResultMeasureValueDataTypes.Flag or TADA.MeasureQualifierCode.Flag.

I also updated the code which identifies Non-Detect results where a non-detect qualifier code has been entered in ResultMeasureValue by pulling the list of non-detect qualifiers and checking for the presence of any non-detect codes in ResultMeasureValue.


R/CensoredDataSuite.R
```diff
@@ -37,12 +37,21 @@ TADA_IDCensoredData <- function(.data) {
     "TADA.ResultMeasureValueDataTypes.Flag"
   )
   TADA_CheckColumns(.data, expected_cols)
-
-  ## First step: identify censored data
-  cens <- .data %>% dplyr::filter(TADA.ResultMeasureValueDataTypes.Flag == "Result Value/Unit Copied from Detection Limit")
-  not_cens <- .data %>% dplyr::filter(!ResultIdentifier %in% cens$ResultIdentifier)
+  
+  ## Run TADA_FlagMeasureQualifierCode to add column TADA.MeasureQualifier.Flag to allow for using user-supplied Result Measure Qualifier codes to identify censored samples.
+ data_mq_flag <- TADA_FlagMeasureQualifierCode(.data)
```
cristinamullin, line 42: Do you think TADA_FlagMeasureQualifierCode should always run within this function, or should it only run if the TADA.MeasureQualifier.Flag column does not yet exist in the input df?

hillarymarler: I think it makes sense to run the function only if the TADA.MeasureQualifier.Flag column does not yet exist.

cristinamullin: I'll work on this

cristinamullin:
lines 66-94 in TADA_IDCensoredData need more work (ran out of time). commenting out for now

PR #371, opened 2023-12-21
Builds consistency

cristinamullin:
use "NA - Not Applicable" throughout instead of "Blank" in TADA flag fields when NA. Add column ordering to TADA_FlagMeasureQualifierCode. Update refs and example data. Use Mod 1 vignette to include information about qual code def generation.


hillarymarler (approved):
These changes make sense to me in terms of building consistency.

I like the addition to Mod 1 to explain the addition of TADA.MeasureQualifierCode.Def

PR #373, opened 2023-12-28
Update Filtering.R

hillarymarler:
Replaced MeasureQualifierCode with TADA.MeasureQualifierCode.Ref when display = "key".

Added TADA.MeasureQualifierCode.Def when display = "most"

Changes made so TADA.MeasureQualifierCode.Ref will be displayed as a filter option in TADA Shiny


hillarymarler:
@cristinamullin - I am not sure why some of the checks are failing. Any suggestions on what I should try to resolve this?

cristinamullin (approved):

PR #377, opened 2024-01-03
336 original vs tada columns copies

hillarymarler:
1. New function in utilities called "TADA_RetainRequired" which retains only the TADA version of duplicate columns
2. Removed Narrow profile (source of MethodSpecificationName) from Data Retrieval (and related references in examples and documentation) 
3. Changed column names in csv files to reference TADA.MethodSpeciationName where necessary
4.  Edited all functions and documentation to instead refer to MethodSpeciationName from FullPhysicalChem profile
5. Checked that workflow still ran smoothly by running through TADA Module 1 after changes were made


cristinamullin:
@hillarymarler I made a few changes: 
1) exported the new function TADA_RetainRequired, edited it to return .data, and added an example. It is best practice to always include examples (some of the functions we created early on still do not have them, but we're still working on that)
2) unrelated to your edits: I removed two of the example files that did not have good documentation on how they were generated. Instead of including new example files, I updated the example for the join function.  I also found an issue with the USGS values in the activity type ref, which I corrected as part of the PR as well

cristinamullin (approved):

PR #386, opened 2024-01-17
149 address depth profile data

hillarymarler:
Function to identify depth categories and determine mean, max, or min (defined by user input) by either monitoring location/date or monitoring location/date/depth category.

Should there be a "clean" option which would return only the aggregate values for groups with n > 1?


R/ResultFlagsIndependent.R
```diff
@@ -1254,3 +1254,307 @@ TADA_FindPotentialDuplicatesSingleOrg <- function(.data) {
...
+#' bottom= "Hypolimnion-bottom", and the full depth profile for
+#' "Metalimnion/Thermocline-middle".
+#'
+#' When more than one result is available for a MonitoringLocationIdentifier,
+#' ActivityStartDate, OrganizationIdentifier, and TADA.CharacteristicName,
+#' user input defines which value (average, max, or min value) to use for that
+#' day and location. User input also defines whether aggregation occurs over
+#' each depth category or for the entire depth profile.
+#'
+#' @param .data TADA dataframe which must include the columns ADD DEPTH COLUMNS
```
cristinamullin, line 1273: add depth columns

R/ResultFlagsIndependent.R
```diff
@@ -1254,3 +1254,307 @@ TADA_FindPotentialDuplicatesSingleOrg <- function(.data) {
...
+#' combination. An additional column, TADA.ResultValueAggregation.DepthProfile.Flag will be added
+#' to describe aggregation.
+#'
+#' @param bycategory Boolean argument with options "TRUE" or "FALSE". The default is
+#' by category = "FALSE" which means that any descriptive statistics calculated are
+#' based on the entire water column at a Monitoring Location. When bycategory = "TRUE",
+#' any calculated descriptive statistics are determined for each depth category for each
+#' Monitoring Location.
+#'
+#' @param clean Boolean argument with options "TRUE" or "FALSE". The default is
```
cristinamullin, line 1292: "clean" or "aggregatedonly"? 

We use clean in other other function to remove data that is invalid or not applicable to the dataset. In this situation for depth data, clean might be more applicable to use to remove results that do not include depth data (I wonder if that should be it's own separate function - a flag function for finding results that do not include any depth information). 

aggregatedonly is similar to other flag functions flaggedonly function input.

R/ResultFlagsIndependent.R
```diff
@@ -1254,3 +1254,307 @@ TADA_FindPotentialDuplicatesSingleOrg <- function(.data) {
...
+      # determine the number of Depths per group
+      dplyr::mutate(
+        DepthsPerGroup = length(unique(Depth)),
+        # determine bottom value using TADA.ActivityBottomDepthHeightMeasure.MeasureValue or the max depth record for profile data
+        Bottom = ifelse(DepthsPerGroup > 1 & is.na(TADA.ActivityBottomDepthHeightMeasure.MeasureValue), max(Depth), TADA.ActivityBottomDepthHeightMeasure.MeasureValue)
+      ) %>%
+      dplyr::ungroup() %>%
+      # assign depth categories by using depth information
+      dplyr::mutate(TADA.DepthCategory.Flag = dplyr::case_when(
+        !MonitoringLocationTypeName %in% c("Lake, Reservoir, Impoundment", "Great Lakes", "Lake", "Great Lake") ~ "Not Calculated For MonitoringLocationType",
```
cristinamullin, line 1353: Does this function only work for these location types?
"Lake, Reservoir, Impoundment", "Great Lakes", "Lake", "Great Lake"

Suggest to add details about this to function documentation.

R/ResultFlagsIndependent.R
```diff
@@ -1254,3 +1254,307 @@ TADA_FindPotentialDuplicatesSingleOrg <- function(.data) {
...
+#'
+#' @param daily_agg Character argument; with options "no", "ave", "min", or
+#' "max". The default is daily_agg = "no". When daily_agg = "no", all results
+#' will be retained. When daily_agg == "ave", the mean value in each group of
+#' results (as determined by the by category param) will be determined for each
+#' MonitoringLocation, ActivityDate, Organization ID, and TADA.CharacteristicName combination.
+#' When daily_agg == "min" or when daily_agg == "max", the min or max
+#' value in each group of results (as determined by the by category param) will
+#' be determined for each MonitoringLocation, ActivityDate, and TADA.CharacteristicName
+#' combination. An additional column, TADA.ResultValueAggregation.DepthProfile.Flag will be added
```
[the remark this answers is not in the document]
hillarymarler: I like that - will make the  change.

R/ResultFlagsIndependent.R
```diff
@@ -1254,3 +1254,307 @@ TADA_FindPotentialDuplicatesSingleOrg <- function(.data) {
...
+#' to describe aggregation.
+#'
+#' @param bycategory Boolean argument with options "TRUE" or "FALSE". The default is
```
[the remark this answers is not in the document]
hillarymarler: Would "all" combine all categories? Or give a separate aggregate value for each category?

R/ResultFlagsIndependent.R
hillarymarler: Thanks for the explanation. I will change clean to "aggregatedonly" for this function. I think a "clean" option for this function also makes sense.

hillarymarler:
@cristinamullin  - I made the changes you suggested

hillarymarler:
Need to:
1) add to Module 1
2) update ref

hillarymarler:
I am currently working on the depth profile figure function as part of this pull request.

hillarymarler:
In addition to the depth profile figure and TADA_IDDepthProfiles function, I also changed the depth category flags to "Surface", "Middle", and "Bottom" as we discussed.

hillarymarler:
I removed id_cols as an option, so groups to be plotted must be selected from TADA.ComparableDataIdentifier. If TADA.ComparableDataIdentitifer column is not in data set, the function will stop and provide a message to user to run TADA_CreateComparableID

cristinamullin:
``` r
❯ checking examples ... [347s] ERROR
  Running examples in 'TADA-Ex.R' failed
  The error most likely occurred in:
  
  > base::assign(".ptime", proc.time(), pos = "CheckExEnv")
  > ### Name: TADA_DepthProfilePlot
  > ### Title: Create A Three-Characteristic Depth Profile
  > ### Aliases: TADA_DepthProfilePlot
  > 
  > ### ** Examples
  > 
  > # Load example dataset:
  > data(Data_6Tribes_5y)
  > # Create a depth profile figure with three parameters for a single monitoring location and date
  > TADA_DepthProfilePlot(Data_6Tribes_5y, 
  +                       groups = c('TEMPERATURE, WATER_NA_NA_DEG C', 'PH_NA_NA_NA', 'DEPTH, SECCHI DISK DEPTH_NA_NA_M'),
  +                       location = "REDLAKE_WQX-ANKE",
  +                       activity_date = "2018-10-04")
  [1] "TADA_DepthProfilePlot: Running TADA_DepthCategoryFlag function to add required columns to data frame"
  [1] "TADA_DepthCategory.Flag: checking data set for depth values. 68903 results have depth values available."
  [1] "TADA_DepthCategory.Flag: assigning depth categories."
  [1] "TADA_DepthCategory.Flag: Grouping results by MonitoringLocationIdentifier, OrganizationIdentifier, CharacteristicName, and ActivityStartDate for aggregation for entire water column."
  [1] "TADA_DepthCategory.Flag: No aggregation performed."
  [1] "TADA_DepthProfilePlot: Identifying available depth profile data."
  Error in `stringr::str_detect()`:
  ! Can't recycle `string` (size 3) to match `pattern` (size 2).
  Backtrace:
      ▆
   1. ├─TADA::TADA_DepthProfilePlot(...)
   2. │ └─stringr::str_detect(groups, depth.params.groups)
   3. │   └─stringr:::check_lengths(string, pattern)
   4. │     └─vctrs::vec_size_common(...)
   5. └─vctrs::stop_incompatible_size(...)
   6.   └─vctrs:::stop_incompatible(...)
   7.     └─vctrs:::stop_vctrs(...)
   8.       └─rlang::abort(message, class = c(class, "vctrs_error"), ..., call = call)
  Execution halted
```

hillarymarler:
I should be able to take a look at this today/tomorrow.

hillarymarler:
We need to standardize units for Secchi depth. I think that at one point I had set this up to happen in TADA_ConvertResultUnits() but that is not happening anymore. It probably isn't the best place anyway as if I'm remembering correctly, that solution meant that all length units were standardized to m which may not be appropriate in every case.

Where would be a better place to standardize Secchi units? I thought about putting it in the depth profile figure function itself, but that is problematic because it means that when users select parameters by TADA.ComparableDataIdentifier, they will be excluding Secchi results if there are results in multiple units.

Would TADA_ConvertDepthUnits or TADA_HarmonizeSynonyms be the better place?

cristinamullin:
It looks like TADA_ConvertResultUnits() is still not working as expected. There are units in our reference table (WQXunitRef) with different target units. With the current implementation, all length units should be standardized to the same unit within that category. Which I agree which may not be appropriate for every characteristic. To implement characteristic specific units, we started to discuss a new approach but it has not been implemented: https://github.com/USEPA/TADA/issues/398

- Removing unit conversion columns from harmonizationtemplate.csv
- Including target units for specific characteristics as a csv input to unit conversion function instead. This would allow users to edit it if needed. It would potentially also let users pick units in the shiny app.

hillarymarler:
In the current WQX unit ref table, it looks like both m and ft are listed as target units. I think removing one or the other (remove ft?), so that only one length unit is listed as the target would solve the problem in the short term and help fix the issues I'm having with the depth profile functions. Should I give that a try? 

Or I could work on the new approach to unit conversion and revisit the depth profile functions after. What is your preference?

hillarymarler:
@cristinamullin I updated this branch to be consistent with develop. I will start working on this again once we've wrapped up the unit conversion work.

PR #388, opened 2024-01-18
Create required columns list and order

cristinamullin:


cristinamullin:
@hillarymarler can you review these lists and their order? Feel free to make edits if you notice anything is missing or the order needs adjustment. Once we're done with this, I can look into using the lists in these functions so we only have to keep one list up to date: TADA_CheckRequiredFields, TADA_OrderCols, TADA_AutoFilter, TADA_RetainRequired, and TADA_GetTemplate

R/RequiredCols.R
```diff
@@ -0,0 +1,231 @@
+# Lists can be used within TADA_OrderCols, TADA_GetTemplate, TADA_CheckRequiredFields, TADA_AutoFilter, TADA_RetainRequired
+
+# ordered list of TADA workflow required columns to be retained in data frame
+require.cols <- c(
```
hillarymarler, line 4: Would it make sense to store the column names in a csv file? An additional column could store which "group" they are in (Result Value, Detection Limits, etc.)? Others could specify which are generated and which are required.

It might be useful to be able to easily specify subsets of the whole list in current or future functions.

R/Utilities.R
```diff
@@ -1275,104 +1286,109 @@ TADA_AutoFilter <- function(.data) {
...
-                "ResultIdentifier",
-                "OrganizationIdentifier")
-  
-  # create list of columns containing all NA values. 
-  na.cols <- .data %>% purrr::keep(~all(is.na(.x))) %>%
-   names()
-  
+
+  # create list of required columns that must be retained even if all values are NA
+  req.cols <- c(
```
hillarymarler, line 1294: For example, this could be a place to just read in the list of required columns from the csv.

R/Utilities.R
```diff
@@ -1387,148 +1403,149 @@ end <- dim(.data)[1]
...
 #' @export
-#' 
-#' @examples 
+#'
+#' @examples
 #' data(Data_Nutrients_UT)
 #' reducedcols_Data_Nutrients_UT <- TADA_RetainRequired(Data_Nutrients_UT)
 #'
-
 TADA_RetainRequired <- function(.data) {
```
hillarymarler, line 1421: Another place where filter/required columns could be imported from csv

hillarymarler (approved):

R/RequiredCols.R
```diff
@@ -0,0 +1,231 @@
...
+  # Comparable Data Groups (e.g. Observable Properties)
```
hillarymarler, line 5: List and order look good to me.

cristinamullin:
My last commit updates all functions that include a required col lists to use this master list. When reviewing, I just noticed TADA_AutoFilter does not exclude an columns in the example included in the documentation. @hillarymarler can you double check that one is still working as expected? Thanks!

hillarymarler:
Bug in TADA_AutoFilter() is fixed now

PR #391, opened 2024-01-22
372 create function to remove groundwater results

hillarymarler:
Function to filter for assessments. Right now it works for surface water, ground water, and sediment. Still need to add fish tissue (once new profiles are available)


R/Filtering.R
```diff
@@ -240,3 +240,161 @@ TADA_FieldValuesTable <- function(.data, field = "null", characteristicName = "n
   dat <- dat %>% dplyr::arrange(desc(Count))
   return(dat)
 }
+
+#' TADA_AssessmentDataFilter
+#'
+#' This function will filter the data set and retain only the media types
+#' selected by the user. It uses ConstructionDateText,
```
cristinamullin, line 247: Add ActivityMediaSubdivisionName, AquiferName, LocalAqfrName

R/Filtering.R
```diff
@@ -240,3 +240,161 @@ TADA_FieldValuesTable <- function(.data, field = "null", characteristicName = "n
...
+    gr.water.data <- .data %>%
+      dplyr::filter(TADA.Groundwater.Flag == "Groundwater") %>%
+      dplyr::mutate(TADA.UseForAssessment.Flag = "Yes")
+    
+  }
+  
+  if (ground_water == FALSE) {
+    gr.water.data <- .data %>%
+      dplyr::filter(TADA.Groundwater.Flag == "Groundwater") %>%
+      dplyr::mutate(TADA.UseForAssessment.Flag = "No")
```
cristinamullin, line 356: Instead of creating separate TADA.Groundwater.Flag and sediment flag columns, can we include this information in the TADA.UseForAssessment.Flag?

Values in the TADA.UseForAssessment.Flag could be more specific than Yes/No to consolidate these extra columns. For example, if the row is GW, then the TADA.UseForAssessment.Flag could be "Groundwater" or "No-Groundwater". 

Same for sediment and surface water ("No-Sediment" or "Yes-Surfacewater".)

cristinamullin:
@hillarymarler Can you add this to the Module 1 vignette?

cristinamullin (approved):

cristinamullin:
Merging in synonym harmonization function fix here to we can push one update to develop for both

PR #395, opened 2024-02-01
Fix TADA_HarmonizeSynonyms issue

cristinamullin:


hillarymarler:
I was able to run the WIDNR example (had to substitute the site id provided in line 4 in TADA_DataRetrieval() as object 'run_details' was not found) and with your changes, the number of rows with non-NA values in TADA.ResultMeasureValue was the same in the harmonized and non-harmonized data sets.

Issue #346, opened 2023-10-06
Add tribal boundaries to map function


cristinamullin:
**Is your feature request related to a problem? Please describe.**

Add tribal reservation boundaries to the [TADA_OverviewMap()](https://usepa.github.io/TADA/reference/TADA_OverviewMap.html).

cristinamullin:
Leverage EPA web services so updates are incorporated automatically: https://www.epa.gov/data/tribes-services

PR #405, opened 2024-02-22
#346 - Add tribal boundaries to overview map 

renaemyers:
Tribal layers will load by default in the overview map, restricted by the bounding box of the current dataset. They can be toggled on and off using a button (all layers work together and can't be turned on/off individually). Colors and icons are as discussed previously (orange/tan colors and open triangle icons for points) but can be changed to match HMW if desired.


cristinamullin:
@renaemyers In this example, if you try to click on the sites, you can't unless the tribal boundary is turned off first.  Is it possible for the WQP monitoring locations to be on top of the tribal layer by default?

data(Data_6Tribes_5y_Harmonized)
TADA_OverviewMap(Data_6Tribes_5y_Harmonized)

cristinamullin:
Updates: 
- Updated function names to match TADA convention and exported: TADA_addPoints and TADA_addPolys. I think they could be values to others on their own.
- Added TADA::: prefix to any examples in documentation that call functions that are NOT exported.
- Ran document()
- Merged in updates from develop upstream

PR #413, opened 2024-03-04
410 whole numbers for characteristic and measurement legends in tada overviewmap

hillarymarler:
colorBins uses the pretty() function to create bins. I used round() with pretty() outside of colorBin to specify that breaks should be whole numbers only. This prevents any decimals from being used.

This does do weird things when data sets have small values for Parameter_Count. For example for the Utah nutrients data set, it just creates one bin (1-2). I can specify that "0" should be added to the vector for bins, but then it would create the categories "0-1", and "1-2" and then display all samples with the "1-2" color so it does not make much sense.

We could consider doing something like checking for the max Parameter_Count (maybe n < 5) and if it is under a certain number specify the colors to match a particular integer with colorFactor instead of colorBin to create the palette. Maybe it would be important to also check to make sure that 0 has not been added as a break by pretty() before switching to a colorFactor approach? I'm not sure sure yet.

I can keep exploring it , but wasn't sure how important being able to identify between very small Parameter_Count values is.


cristinamullin:
@hillarymarler I just commented on this linked issue as well. I like your idea of having specific colors for integers 1-5 (no including 0) if all sites have 5 or less parameters. Another option might be to not color the sits based on N parameters if there are less than 5 but it may be important to some organizations (e.g. tribes) that are only monitoring a few parameters at each site.

cristinamullin (approved):

PR #420, opened 2024-03-11
375 range percent values 26 50 not handled correctly in tada convertspecialchars

hillarymarler:


R/Utilities.R
```diff
@@ -299,6 +299,10 @@ TADA_CheckColumns <- function(.data, expected_cols) {
 #' @param .data A TADA profile object
 #' @param col A character column to be converted to numeric
 #'
+#' @param percent.ave Boolean argument; default is clean = TRUE. When clean = TRUE,
```
cristinamullin, line 302: The clean argument mentioned here in the documentation should be percent.ave

hillarymarler:
@cristinamullin I fixed the documentation error.

hillarymarler:
@cristinamullin - I will review this again and run some more examples tomorrow morning.

hillarymarler:
@cristinamullin  - this is ready for you to review

cristinamullin:
the DO full join is causing tada DR to print the join columns
image

We could use suppressMessages but if possible it is better to specify the "by" argument to avoid printing this message. dplyr only prints a message to let you know what its guess is for which columns to join by. In this case, we are joining by all columns. Adding a commit now to fix.

cristinamullin (approved):

PR #430, opened 2024-03-25
tribal layers always load under circle markers

renaemyers:


R/Utilities.R
```diff
@@ -1021,12 +1020,10 @@ TADA_addPolys <- function(map, url, layergroup, layername, bbox = NULL) {
 #' @param bbox A bounding box from the sf function st_bbox; used to filter the query results. Optional; defaults to NULL.
 #' @return The original map with polygon from the feature layer added to it.
 #'
-#' @export
-#'
 #' @examples
 #' \dontrun{
 #' # Create a leaflet map
-#' lmap <- leaflet::leaflet() %>% leaflet::addProviderTiles("Esri.WorldTopoMap", group = "World topo")
+#' lmap <- leaflet::leaflet() %>% leaflet::addProviderTiles("Esri.WorldTopoMap", group = "World topo")  %>% leaflet::addMapPane("tribes", zIndex = 300)
```
cristinamullin, line 1026: The examples stopped working anymore for TADA_addPolys and TADA_addPoints. I fixed this but also removed these two functions as their own exports for now. Do you think we should use "tribes" here, options = leaflet::pathOptions(pane = "tribes"), or would it make more sense to use something more generic if these utilities may be used to add other poly or point layers to the map as well?

PR #431, opened 2024-03-27
334 flag duplicates for multiple orgs weird behavior to check

hillarymarler:
New version of TADA_FindPotentialDuplicatesMultipleOrgs whichlooks for NearbySites first, subsets the data and then looks for matching result values, activity date/time. This has the result of all group IDs for organizations or nearby sites increasing incrementally (no skipped #s in the final data set).

I checked a few data sets against the original function and the only minor difference I saw was that in one case where the same organization had a duplicate result from a nearby site, the old vs. new version might select a different result as that selection happens as the result of dplyr::slice_sample(n=1) (random selection) after the result options have been filtered to the lowest org rank.

I also tried writing a few tests for this function to check and make sure that group IDs are incremental (or NA in the case of no duplicates detected) and that all new columns added by the function contain a value in each row.


cristinamullin (approved):

cristinamullin:
@hillarymarler this looks great. I am slipping in a test for autoclean into this PR since we have a join within that now. I also made a few small updates to the docs for the duplicate function. Since the example runs TADA_DataRetrieval I added dontrun so the example won't run as part pf the checks.

PR #434, opened 2024-04-03
Progress towards linking WQP to ATTAINS data

kathryn-willi:
Hi Cristina,

This PR works towards completing the TADA goal: _Match Water Quality Portal (WQP) sites with ATTAINS assessment units. Assist user with reviewing sites and Assessment Units (AUs) on a map and deciding to keep or exclude sites (systematically first, maybe using catchments as a buffer around the AU points/line/polygons for consistency, and then let users make manual edits to include or exclude sites (main requirement)). Consider adding option to create new AUs for new areas where WQP sites are available but no AU exists yet (secondary requirement). Output is a simple spreadsheet linking ATTAINS AU IDs with WQP Location IDs._

Though what we are submitting here does not exactly close/fulfill an issue in {TADA}, we have made headway towards addressing the goal above by developing the following 4 functions:

**TADA_MakeSpatial()**: converts a TADA-style Water Quality Portal data frame into a spatial {sf} object.

**TADA_GetATTAINS()**: links WQP data to the ATTAINS assessment unit features that intersect the same NHD High Resolution catchments as them.

**TADA_ViewATTAINS()**: maps a summary of the same data pulled from TADA_GetATTAINS().

**fetchATTAINS()**: an intermediate function for pulling ATTAINS data based on a bounding box generated by a user-supplied set of WQP observations.

This PR also creates a new vignette outlining these functions (excluding **fetchATTAINS()** - which is still not an "exported" function for TADA), and a {testthat} suite of basic tests we've performed on them.

These new functions are documented using roxygen2 syntax and the documentation files have been updated with devtools::document(). Dependencies, warnings, and errors were checked with devtools::check() and devtools::test().

I put all of these functions in a single .R script and named it what I named the associated vignette and {testthat} file - "GeospatialFunctions". I figure this would be a simple solution for keeping all of our ROSSyndicate functions in a place that's a little separated from the rest of TADA, at least for now!

Beyond your code review, I am hoping you could comment on a couple things my team's been thinking about:

As mentioned previously, these functions are not exactly what the end-users wanted, but they are definitely progress towards it. Moreover, they are still in development and we have ideas for improving them. In the future, how should we handle these sorts of "stepping stone" functions? Should we wait to send them to you until all needs have been fully addressed?

Thanks for the opportunity to contribute, and I look forward to your feedback and learning how best to submit future PRs! 😄 

-Katie


R/GeospatialFunctions.R
```diff
@@ -0,0 +1,611 @@
...
+    purrr::map(~ feature_downloader(.))
+  
+  names(final_features) <- c("ATTAINS_catchments", "ATTAINS_points", "ATTAINS_lines", "ATTAINS_polygons")
+  
+  return(final_features)
+  
+}
+
+
+#' TADA_GetATTAINS
```
hillarymarler, line 211: I am getting some inconsistent results when using TADA_GetATTAINS. With small TADA data frames over a smaller area (like a HUC), all the expected geometry (catchment outlines, polyons, lines, points) is returned and appears on the map with TADA_ViewATTAINS. For larger data sets across a larger area, sometimes one or more of the ATTAINS objects expected are returned with a NULL value so then don't show up with TADA_ViewATTAINS.

This is an example of a data set where I saw those issues:
test1 <- TADA_DataRetrieval(startDate = "2018-05-01", 
                           endDate = "2018-09-30", 
                           statecode = "IL", 
                           applyautoclean = TRUE)

There should be both lines and polygons, but only polygons are successfully included in the output.

Sometimes when I run the above example (or other similar ones) in fetchATTAINS, I will get an error message with a URL. At the URL, the final lines visible indicate: "exceededTransferLimit": true. 
image

R/GeospatialFunctions.R
```diff
@@ -0,0 +1,611 @@
...
+                "ATTAINS_catchments" = ATTAINS_catchments,
+                "ATTAINS_points" = ATTAINS_points,
+                "ATTAINS_lines" = ATTAINS_lines,
+                "ATTAINS_polygons" = ATTAINS_polygons))
+  }
+  
+}
+
+
+#' TADA_ViewATTAINS
```
hillarymarler, line 422: I really like the appearance of the features in the map. It is easy to identify the lines/polygons/points vs the catchment outlines relative to the MonitoringLocations. I am wondering if long term it would make sense to pull catchment outlines from a source other than ATTAINS because my understanding is that only catchments with associated Assessment Units are available from the ATTAINS webservices. Being able to see catchment outlines in areas where an org has not yet created an assessment unit (but has new MonitoringLocations) might be very useful when making determinations about which MonitoringLocations should be included within new assessment units.

hillarymarler (commented):
Overall, I'm really excited about these new functions! They will be so helpful for TADA users. I had trouble getting/viewing all ATTAINS objects that I expected for particular data sets (more details, with an example in other comments).

R/GeospatialFunctions.R
```diff
@@ -0,0 +1,611 @@
...
+#'                                        characteristicName = "pH",
+#'                                        statecode = "SC", 
+#'                                        countycode = "Abbeville",
+#'                                        applyautoclean = TRUE)
+#' 
+#' # make `tada_not_spatial` an {sf} object, projected in crs = 4269 (NAD83)                                 
+#' tada_spatial <- TADA_MakeSpatial(data = tada_not_spatial, crs = 4269)
+#' }
+
+TADA_MakeSpatial <- function(data, crs = 4326){
```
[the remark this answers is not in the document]
cristinamullin: FYI - I'm working on switching this to use .data now.

R/GeospatialFunctions.R
cristinamullin: @kathryn-willi is this something you and you team have time to look into before we merge this PR in? Or would you prefer we create a separate issue for this and address it in a separate PR later?

R/GeospatialFunctions.R
cristinamullin: I agree - "Being able to see catchment outlines in areas where an org has not yet created an assessment unit (but has new MonitoringLocations) might be very useful when making determinations about which MonitoringLocations should be included within new assessment units.". 

I'll write up a separate issue to capture this, and we can discuss details with the team (e.g. what additional sources for catchment we would want to use and how they match up/don't match up with the catchments ATTAINS uses) as a potential future improvement. For example, I believe the ATTAINS catchments are a mix of NHD high res catchments and hexagons for some oceans/lakes/estuaries. I am not sure how that will mesh visually with if we bring in a different catchment layer for the areas that the ATTAINS catchment layer does not cover.

R/GeospatialFunctions.R
kathryn-willi: Hi Cristina! We have been working on fixing the issue, but we haven't been able to reproduce the error. 🤔 I am happy to continue troubleshooting ahead of merging this PR, but I also do not want this to slow down progress on your other work. If you are comfortable with merging as-is and us creating an issue, we will continue to work towards addressing it!

cristinamullin: @hillarymarler could you share a more detailed reprex to help Katie reproduce the errors you ran into/check if you are still running into them now? I wonder if the ATTAINS API time out issues are intermittent (might not be reproducible), similarly to the WQP.

cristinamullin (approved):

cristinamullin:
@kathryn-willi regarding your questions: "As mentioned previously, these functions are not exactly what the end-users wanted, but they are definitely progress towards it. Moreover, they are still in development and we have ideas for improving them. In the future, how should we handle these sorts of "stepping stone" functions? Should we wait to send them to you until all needs have been fully addressed?"

I think what you have in this PR is perfect - no need to wait. This is great progress, and it's worth sharing/merging in at this point even though updates are coming. It is helpful to be able to review and provide feedback along the way in case they impact future improvement/dev work related to these functions - for example, you'll see some of the changes we made for consistency with the TADA package when you merge them back into your fork). Future improvements can be made in subsequent PR's. 

Really great work - thank you again!

PR #448, opened 2024-04-30
398 remove unit conversions from tada harmonization template

hillarymarler:
Unit conversions can be customized in TADA_ConvertResultUnits.

The new function TADA_CreateUnitRef is used to create a reference df based on TADA priority characteristics (then WQX/USGS targets). This function is run within TADA_ConvertResultUnits if "tada" is selected as the ref. It can also be run by the user to create a draft unit ref that can be customized for use within TADA_ConvertResultUnits.

Temperature conversions now happen along with all other conversions.


hillarymarler:
I'm taking a look at the test failure issues.

hillarymarler:
Still have work to do regarding harmonization tests.

hillarymarler:
@cristinamullin - this is ready for your review

PR #470, opened 2024-05-31
Trainingupdates

hillarymarler:
Fix for some depth profile plot issues (now it will display depth categories correctly). It still is not displaying on the pkgdown site, but it will run and display correctly when running the code chunks in RStudio.


cristinamullin:
@hillarymarler I noticed an issue with the WQX validation table. Some values with the nonstandardized flag were being flagged as not reviewed. I fixed this, and also updated the internal ref files and example datasets.

cristinamullin (approved):

PR #480, opened 2024-06-10
473 tada twocharacteristicscatterplot legend should not include fraction andor speciation if they are na

hillarymarler:
Fixed this for legend and y axis. I think a long term solution that could be applied here and other places is to create an internal function that uses stringr::str_remove_all with common iterations of NA found in TADA.ComparableDataIdentifier, etc. that could be used in many of the current and future figure functions. 

If you think this sounds like a good idea, @cristinamullin , I can write up an issue for it.


cristinamullin:
Yes, good idea!

cristinamullin (approved):

PR #486, opened 2024-06-17
471 tada convertresultunits is overwriting tadamethodspeciationname with na

hillarymarler:
Updates to fix overwrite of tada method speciation name with na. In order to get this to work correctly, I had to add TADA.MethodSpeciationName as a column in the output of TADA_CreateUnitRef.


hillarymarler:
@wokenny13 - if you have time to take a look at these changes and run TADA_ConvertResultUnits on a few example data sets that would be great. Thanks!

hillarymarler:
@wokenny13 - this is ready for review.

wokenny13:
Ran through a few example datasets for different states and for both Nitrogen and Phosphorus. The updates have fixed the overwriting issues for these examples that I ran.

hillarymarler:
@wokenny13 - thanks for running some example data sets. The checks are re-running due to a new vignette I merged into the develop branch. Once all those checks have passed, I will merge in the unit conversion changes.

PR #490, opened 2024-07-01
485 tadacomparabledataidentifier is used for legend labels in tada twocharacteristicscatterplot uses a different column for id cols

wokenny13:
Current progress on addressing the legend labels in TADA_TwoCharacteristicScatterplot. See previous comments for details.


hillarymarler:
@wokenny13 - it looks like there are some check failures related to use of this function in at least one vignette? Do you want to try and troubleshoot that before I review? Or do you want me to look into it/correct it as part of my review?

hillarymarler:
@wokenny13 - I can review the function updates while you are working on check issues related to vignette(s). I'll take a look either later this afternoon or first thing tomorrow morning and we can discuss during our working session tomorrow.

R/Figures.R
```diff
@@ -841,7 +841,7 @@ TADA_Scatterplot <- function(.data, id_cols = c("TADA.ComparableDataIdentifier")
 #' # Creates a scatterplot including the two specified sites in the same plot:
 #' TADA_TwoCharacteristicScatterplot(df, id_cols = "MonitoringLocationName", groups = c("Upper Red Lake: West", "Upper Red Lake: West-Central"))
 #'
-TADA_TwoCharacteristicScatterplot <- function(.data, id_cols = "TADA.ComparableDataIdentifier", groups) {
+TADA_TwoCharacteristicScatterplot <- function(.data, id_cols = "TADA.ComparableDataIdentifier", groups, id_cols2 = NULL, groups2 = NULL) {
```
hillarymarler, line 844: Can the new params have more descriptive names? Should group be characteristic(s) and groups2 by groups?

R/Figures.R
```diff
@@ -841,7 +841,7 @@ TADA_Scatterplot <- function(.data, id_cols = c("TADA.ComparableDataIdentifier")
 #' # Creates a scatterplot including the two specified sites in the same plot:
```
hillarymarler, line 841: New params need to be added in function description. Will there be a list of allowable values for the new params? Or can the user enter any column name as a group?

R/Figures.R
```diff
@@ -864,30 +864,62 @@ TADA_TwoCharacteristicScatterplot <- function(.data, id_cols = "TADA.ComparableD
   }
 
   if (!"TADA.ComparableDataIdentifier" %in% id_cols) {
```
hillarymarler, line 866: Should TADA.ComparableDataIdentifier be required/automatic for id_cols? It makes the most sense to use here (so that characteristic and unit will be comparable for all results). What other options do you think might be used here?

R/Figures.R
```diff
@@ -864,30 +864,62 @@ TADA_TwoCharacteristicScatterplot <- function(.data, id_cols = "TADA.ComparableD
...
+    stop("Note: TADA.ComparableDataIdentifier not found in id_cols argument.")
   }
 
   # check that groups are in id_cols
   id <- unlist(unique(.data[, id_cols]))
   if (any(!groups %in% id)) {
     stop("The 'groups' vector contains one or more inputs that are not found within your input dataset. Check spelling and try again.")
   }
+  
+  # check that groups are length two for comparing two characteristics
```
hillarymarler, line 876: Reword comment? Do the groups need to be two characteristics? Couldn't you be comparing one characteristic at two locations?

R/Figures.R
```diff
@@ -864,30 +864,62 @@ TADA_TwoCharacteristicScatterplot <- function(.data, id_cols = "TADA.ComparableD
...
+  if (length(groups) != 2) {
+    stop("The 'groups' vector does not contain two characteristics to compare.")
```
hillarymarler, line 878: Should this first "groups" param be able to accept only one characteristic? For cases when you want to compare one characteristic at multiple locations?

R/Figures.R
```diff
@@ -864,30 +864,62 @@ TADA_TwoCharacteristicScatterplot <- function(.data, id_cols = "TADA.ComparableD
...
+  # if another grouping column wants to be plotted, check that groups2 are in id_cols2
+  
+  if (!is.null(id_cols2)) {
+    id2 <- unlist(unique(.data[, id_cols2]))
+    if (any(!groups2 %in% id2)) {
+    stop("The 'groups2' vector contains one or more inputs that are not found within your input dataset. Check spelling and try again.")
+    }
+  }
+  
+  # check that groups2 are length two for comparing two groups
```
hillarymarler, line 890: Should the function generate a plot anyway if there is only one group? There wouldn't be a comparison, but it seems like a plot could still be generated? Or should the message prompt the user to use TADA_Scatterplot?

hillarymarler (commented):
From a functionality standpoint, your changes make sense. I'll run a few more examples to see if I come up with any additional questions before we meet to discuss. 

Overall, I think the documentation and param names could be updated to provide clarity for the user. I think we may want to consider some constraints on the columns that users can select for various arguments. For example, limiting id_cols to TADA.CharacteristicName, CharacteristicName, or TADA.ComparableDataIdentifier (or just TADA.ComparableDataIdentifier?). I think we should also consider what should happen if a user provides only one "group".

Great work so far!

R/Figures.R
```diff
@@ -1127,6 +1127,340 @@ TADA_TwoCharacteristicScatterplot <- function(.data, id_cols = "TADA.ComparableD
...
+#'   left axis showing the units of a single characteristic groups
+#'   plotted on the same figure area wit a legend for each groups
+#'
+#' @export
+#'
+#' @examples
+#' # Load example dataset:
+#' data(Data_Nutrients_UT)
+#' # Create a single scatterplot with two monitoring locations (groups) for a single TADA.ComparableDataIdentifier. 
+#' # Dataframe has multiple comparable data identifier that was not filtered, select value for 'NITROGEN_TOTAL_AS N_MG/L' to plot this characteristic value.
```
hillarymarler, line 1164: Example needs to show how to "select value for 'NITROGEN_TOTAL_AS_N_MG/L" as now that the documentation for this function is being created, this example is failing to run.

I was not sure how to select the value because Monitoring Locations were already specified in the groups argument.

image

R/Figures.R
```diff
@@ -1127,6 +1127,340 @@ TADA_TwoCharacteristicScatterplot <- function(.data, id_cols = "TADA.ComparableD
...
+#'   to run this function.
+#'
+#' @param id_cols The column in the dataset used to identify the unique groups to
+#'   be plotted. Defaults to 'TADA.ComparableDataIdentifier'. This input is flexible,
+#'   however, for the specific use case in the TADAShiny app where a user might
+#'   create groups based on a concatenation of the comparable data identifier
+#'   with other additional grouping variables (e.g. site type, site name, year,
+#'   organization, etc.)
+#'
+#' @param groups A vector of two identifiers from the id_cols column. For
```
hillarymarler, line 1148: Does it make sense to use TADA.ComparableDataIdentifier here as example? Or allow it as a function input at all? As the two characteristic scatterplot function seems to already handle this scenario?

R/Figures.R
```diff
@@ -1127,6 +1127,340 @@ TADA_TwoCharacteristicScatterplot <- function(.data, id_cols = "TADA.ComparableD
...
+#'   example, if the id_cols is 'TADA.ComparableDataIdentifier', the groups could
+#'   be 'DISSOLVED OXYGEN (DO)_NA_NA_UG/L' and 'PH_NA_NA_NA'. These groups will
+#'   be specific to your dataset. If the id_cols is 'MonitoringLocationName',
+#'   the groups could be 'Upper Red Lake: West' and 'Upper Red Lake: West-Central'.
```
hillarymarler, line 1152: Maybe using something like a county or state code here as an example in addition to the Monitoring Location examples would work?

R/Figures.R
hillarymarler: Could it be made to run more like TADA_Boxplot where if there is more than one TADA.ComparableDataIdentifier, a list of plots is created?

This is from the Create Boxplot(s) documentation: "The user can include
#'   additional grouping columns in the id_cols input. If more than one group
#'   exists in the data frame (i.e. two or more unique comparable data
#'   identifiers), the function creates a list of plots, where each list element
#'   name is a unique group identifier."

R/Figures.R
hillarymarler, line 1148: Documentation references two groups, but function accommodates four.

R/Figures.R
```diff
@@ -1127,6 +1127,340 @@ TADA_TwoCharacteristicScatterplot <- function(.data, id_cols = "TADA.ComparableD
...
+#' # Dataframe has multiple comparable data identifier that was not filtered, select value for 'NITROGEN_TOTAL_AS N_MG/L' to plot this characteristic value.
+#' TADA_MultiScatterplot(Data_Nutrients_UT, id_cols = c("TADA.ComparableDataIdentifier", "MonitoringLocationName"), groups = c("UTAH LAKE 1 MI EAST OF PELICAN POINT", "Utah Lake 2 miles west of Vineyard"))
+#'
+#' # Load example dataset:
+#' data(Data_6Tribes_5y_Harmonized)
+#' # Filter the example data so it includes only one TADA.ComparableDataIdentifier
+#' df <- dplyr::filter(Data_6Tribes_5y_Harmonized, TADA.ComparableDataIdentifier %in% c("TOTAL PHOSPHORUS, MIXED FORMS_UNFILTERED_AS P_MG/L"))
+#' # Creates a scatterplot of the three specified sites of interest in the same plot.
+#' TADA_MultiScatterplot(df, id_cols = c("TADA.ComparableDataIdentifier", "MonitoringLocationName"), groups = c("Upper Red Lake: West", "Upper Red Lake: West-Central","Upper Red Lake: East Central"))
+#' # If no groups are selected, return the top 4 groups by MonitoringLocationName field counts.
```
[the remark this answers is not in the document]
wokenny13: Yes, this is correct

R/Figures.R
```diff
@@ -1127,6 +1127,340 @@ TADA_TwoCharacteristicScatterplot <- function(.data, id_cols = "TADA.ComparableD
...
+#' TADA_MultiScatterplot(df, id_cols = c("TADA.ComparableDataIdentifier", "MonitoringLocationName"))
+#' # Both commands below will return a single scatterplot for just the single comparable data identifier.
```
hillarymarler, line 1175: Should this be included as an example as it is essentially recreating TADA_Scatterplot? Should users running this function be prompted to use TADA_Scatterplot instead if they don't need groups? Or should we retire TADA_Scatterplot (perhaps changing the name of the new TADA_MultiScatterplot function to TADA_Scatterplot?) as this function can handle that use case as well?

This would be a good thing to discuss at a team meeting before we make a decision either way.

wokenny13: I was thinking the same. It would just provide mainly the same output as TADA_Scatterplot at that point, except with a single legend option. It would be like an alternative function to use besides TADA_Scatterplot

R/Figures.R
hillarymarler: You are definitely on the right track with wanting to give users the ability to choose a single one to plot or to generate multiple plots.

I think generating the list of plot as in TADA_Boxplot, TADA_Scatterplot, and TADA_Histogram is the way to go here too. That will keep things consistent within the basic package figure functions and lend itself more easily to incorporation into a automated work flow, rather than requiring the manual selection of a TADA.ComparableDataIdentifier. The examples from these functions highlight that using a data frame filtered for one TADA.ComparableDataIdentifier will result in one plot, while using an unfiltered data frame will result in a list of plots.

R/Figures.R
hillarymarler: Maybe you can demo the no-group output of your new function and the output of the original TADA_Scatterplot in our team meeting? That way we can all be on the same page to make a decision.

R/Figures.R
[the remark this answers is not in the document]
hillarymarler: I updated the language here to reflect this.

R/Figures.R
wokenny13: Yes, I can demo this at the team meeting for us to decide

R/Figures.R
```diff
@@ -1138,6 +1138,385 @@ TADA_TwoCharacteristicScatterplot <- function(.data, id_cols = "TADA.ComparableD
...
+  )
+  
+  # check .data has required columns
+  TADA_CheckColumns(.data, reqcols)
+  
+  # if left blank, ensure comparable data identifier is in the id_cols vector
+  if (is.null(id_cols) | length(unique(id_cols)) == 1) {
+    id_cols <- c("TADA.ComparableDataIdentifier","TADA.ComparableDataIdentifier")
+  }
+  id_cols <- c("TADA.ComparableDataIdentifier", vecsets::vsetdiff(id_cols, "TADA.ComparableDataIdentifier"))
```
hillarymarler, line 1233: Do you think it would be possible to use a function from a different package (one already used in EPATADA) here? Is there a way that "setdiff" could be used here instead? We already have quite a few dependencies and if this is the only place where this package is used, an alternative would be preferable if possible.

Should users be able to provide an id_cols input that does not include TADA.ComparableDataIdentifier as one of the inputs? If they should not be able to, this section can probably be simplified as users would only have to supply one id_col like MonitoringLocationIdentifier or WaterbodyTypeName.

If TADA.ComparableDataIdentifier should not be a required/predetermined id_cols, what are some examples of other groups of columns that might be used?

R/Figures.R
```diff
@@ -1138,6 +1138,385 @@ TADA_TwoCharacteristicScatterplot <- function(.data, id_cols = "TADA.ComparableD
...
+  
+  # if id_cols are specified by user and the comparable data identifier is not selected.
+  if (!"TADA.ComparableDataIdentifier" %in% id_cols) {
+    stop("Note: TADA.ComparableDataIdentifier not found in id_cols argument.")
```
hillarymarler, line 1237: Should this be a "stop" or a  "print"? If we don't want the function to run if TADA.ComparableIDataIdentifier isn't a group_id, then it can just be automatically included as a grouping column. If we do want users to have some flexibility, for example to use TADA.CharacteristicName, instead of TADA.ComparableDataIdentifier, then use "print" to print a warning message.

R/Figures.R
```diff
@@ -1138,6 +1138,385 @@ TADA_TwoCharacteristicScatterplot <- function(.data, id_cols = "TADA.ComparableD
...
+  }
+  
+  # Only allows for 2 column selections in the id_cols vector, with TADA.ComparableDataIdentifier being a required value
+  if (length(id_cols) > 2) {
+    stop("id_cols argument can only be a maximum of length two.")
+  }
+  
+  # Runs if more than 1 Tada.ComparableDataIdentifiers are found in a dataframe. Prompts for user selection of a  single characteristic in the dataframe
```
hillarymarler, line 1245: Can this commented out code be removed?

R/Figures.R
```diff
@@ -1138,6 +1138,385 @@ TADA_TwoCharacteristicScatterplot <- function(.data, id_cols = "TADA.ComparableD
...
+  # var <- NULL
+  # id2 <- NULL
+  # if(length(unlist(unique(.data[, id_cols[1]]))) > 1){
+  #   var <- utils::menu(unlist(unique(.data[, id_cols[1]])), title = "Please specify Characteristic to plot: ")
+  #   assign("id2", unlist(unique(.data[, id_cols[1]]))[var])
+  # }
+  
+  # if groups are not specified, select the top 4 groups to plot.
```
hillarymarler, line 1253: Add "based on number of field counts" to the comment as well as the print note

R/Figures.R
```diff
@@ -1138,6 +1138,385 @@ TADA_TwoCharacteristicScatterplot <- function(.data, id_cols = "TADA.ComparableD
...
+  if (length(id_cols) == 2 & is.null(groups)) {
+    assign("groups", names((utils::head(sort(table(.data[,id_cols[2]]), decreasing = TRUE), 4))))
+    print("Note: No 'groups' specified. Plotting up to top 4 'groups' based on the number of field counts")
```
hillarymarler, line 1256: Maybe "Plotting up to 4 'groups' based on the greatest number of results"?

Should 'groups' by substituted with pasting in the column name of the grouping variable to provide more information for users?

R/Figures.R
```diff
@@ -1138,6 +1138,385 @@ TADA_TwoCharacteristicScatterplot <- function(.data, id_cols = "TADA.ComparableD
...
+  }
+  
+  # check that groups are in id_cols
+  id <- unlist(unique(.data[, id_cols[2]]))
+  if (any(!groups %in% id)) {
+    stop("The 'groups' vector contains one or more inputs that are not found within your input dataset. Check spelling and try again.")
```
hillarymarler, line 1262: Would it make sense to make this "stop" message more specific by pasting the strings that were not found in the input data set in to the message?

Usually for "stop" type messages (and in many other cases, although I don't know that as a team we've had a specific discussion about this convention yet), we print the name of the function in the message to help users know where thing went wrong. Example: "TADA_MultiScatterplot: The 'groups' vector...."

R/Figures.R
```diff
@@ -1138,6 +1138,385 @@ TADA_TwoCharacteristicScatterplot <- function(.data, id_cols = "TADA.ComparableD
...
+  }
+  
+  depthcols <- names(.data)[grepl("DepthHeightMeasure", names(.data))]
+  depthcols <- depthcols[grepl("TADA.", depthcols)]
+  
+  plot.data <- as.data.frame(.data)
+  
+  # this subset must include all fields included in plot hover below
+  plot.data <- subset(plot.data, plot.data[, id_cols[2]] %in% groups)[, c(id_cols, reqcols, depthcols, "ActivityStartDateTime", "MonitoringLocationName", "TADA.ActivityMediaName", "ActivityMediaSubdivisionName", "ActivityRelativeDepthName", "TADA.CharacteristicName", "TADA.MethodSpeciationName", "TADA.ResultSampleFractionText")]
+  # if(!is.null(var)){
```
hillarymarler, line 1272: Can this be removed? Or a comment added on what it might be used to address at a later date?

R/Figures.R
```diff
@@ -1138,6 +1138,385 @@ TADA_TwoCharacteristicScatterplot <- function(.data, id_cols = "TADA.ComparableD
...
+  
+  # Returns the param groups for plotting. Up to 4 params are defined.
+  param1 <- param2 <- parm3 <- param4 <- NULL
+  for(i in 1:length(unique(groups))) {
+    assign(paste0("param",as.character(i)), subset(plot.data, plot.data[, id_cols[2]] %in% groups[i]))
+  }
+  
+  all_scatterplots <- list()
+  
+  for (i in 1:length(unique(plot.data$TADA.ComparableDataIdentifier))) {
```
hillarymarler, line 1286: Given that so much of this is dependent on TADA.ComparableDataIdentifier, it seems to make the most sense to revise function to only require user to input the non-TADA.ComparableDataIdentifier id_col and its groups.

hillarymarler (changes requested):
I left some more specific comments in-line, but overall:

The output (plots) is looking great. Running a df that contains multiple values in TADA.ComparableDataIdentifier results in a list of plots similar some of simple plotting functions which is really nice! Excellent updates.

There are some issues with NA units showing up in the plot title. See pH example below:
image

We do have an issue for creating a function to remove NAs from strings, mostly for use in titles like this, which I hope to work on next. So we could just add a comment that the new function to remove NAs from strings should be applied here once it has been created.

Because so much of this function is based on using TADA.ComparableDataIdentifier and the documentation specifies that each scatterplot is specified for a single TADA.ComparableDataIdentifier, it would simplify the user experience to require users to input only the grouping variable.

So the input would look like:
` grouped_scatterplots <- TADA_MultiScatterplot(imaginary_df, group_col = MonitoringLocatioName, groups = c("Fake Lake", "Make Believe Stream") `

Or if the user is allowed/prompted to enter two id_cols and they enter something other than "TADA.ComparableDataIdentifier" then the function should be modified to run off that alternate id_col.

So requested changes are to:

1) Either make TADA.ComparableDataIdentifier a predetermined selection for one of the id_cols or modify function so that the user could select another column and the function would still run (and create an example of this).

2) Review printed stop messages and notes and determine if it would be useful to add the function name, paste in strings, etc. to aid users in troubleshooting.

3) Add comment to apply future function to remove NAs from scatterplot title (or remove NAs from scatterplot title).

4) Remove commented out sections of code or add comment about what they may be used for in future function updates.

5) If the functionality needed in line 1233 is still needed after other updates, can a function from base R or an existing dependency be used instead of adding vecsets?

R/Figures.R
wokenny13: This should be possible. I was going to ask what are your thoughts with adding functions from a different package and if too many dependencies of other packages become problematic. It was used to make sure TADA.ComparableDataIdentifier was the first items in a vector, but if we no longer make this a user input for an argument in id_cols we might not even need vsetdiff or an alternative way to handle.

I prefer the idea of a single id_cols argument. I think I was trying to keep it consistent with some of the other plot functions that were using c('TADA.ComparableDataIdentifier', 'another field').

Would OrganizationFormalName be a good example of column groupings?

R/Figures.R
wokenny13: I think changing this to print makes sense

R/Figures.R
wokenny13: Yeah this can be removed

R/Figures.R
wokenny13: I was thinking the same with substituting the 'groups' string with the column name, so I can make that change.

wokenny13:
I made some edits that are ready for review @hillarymarler

wokenny13:
Made a draft function for handling NAs for figure titles within this pull request. 

I have linked that issue to this pull request in here.

hillarymarler:
Thanks for adding that new function and linking the issue. I'll review changes to scatterplot function and NA removal function tomorrow!

wokenny13:
newplot
Do you think this color schema works better when 4 plots are plotted? I changed the color palette for parmC and parm D to try to make it more visually pleasing. I haven't pushed this commit though so you can view what it currently looks like and let me know your thoughts

hillarymarler:
I like that color update!

wokenny13:
I've pushed the small changes through, with editing some of the coloring and remove the rm(depthcols) line. The checks have gone through. 

- I will keep a note, or post an issue, to consider adding a figure title for the legends in a future update for TADA.GroupedScatter. And take note to address potential TADA color palette consistency pairings.

Issue #472, opened 2024-06-03
speciation included in units


cristinamullin:
**Describe the bug**

This issue may or may not be related to issue https://github.com/USEPA/TADA/issues/471. I believe it is caused by 
[TADA_ConvertResultUnits()](https://usepa.github.io/TADA/reference/TADA_ConvertResultUnits.html). Speciation in result units should be removed from the unit and added to the speciation field. The issue is that the Estimated Unit which is based on the TADA.ResultMeasure.MeasureUnitCode includes speciation. TADA.ResultMeasure.MeasureUnitCode should never include speciation.

See Shiny output on censored data tab: 
![image](https://github.com/USEPA/TADA/assets/46969696/ef33260f-8843-49bd-93a9-750b7ba6d21e)
This was produced when running through the Shepherdstown vignette.

The issue is that the Estimated Unit which is based on the TADA.ResultMeasure.MeasureUnitCode includes speciation. TADA.ResultMeasure.MeasureUnitCode should never include speciation.

In this example, the original ResultMeasure.MeasureUnitCode is blank and the original DetectionQuantitationLimitMeasure.MeasureUnitCode includes the unit with the speciation (coming from USGS NWIS). 

TADA_ConvertResultUnits should copy the DetectionQuantitationLimitMeasure.MeasureUnitCode to the TADA.ResultMeasure.MeasureUnitCode but remove the speciation information from the TADA.ResultMeasure.MeasureUnitCode using USGS_units_speciation.csv and move that information (as N, as P, etc.) to TADA.MethodSpeciationName.

PR #493, opened 2024-07-02
472 speciation included in units

hillarymarler:
Corrected bug - speciation no longer appearing in TADA method detection limit units


hillarymarler:
Details on the original issue here: https://github.com/USEPA/EPATADA/issues/472

Thanks for reviewing - as part of your review, please test TADA_AutoClean and TADA_ConvertResultsUnits with some random data sets to make sure that no additional rows are added and that TADA result unit columns contain values that make sense.

cefergus (commented):
I think the changes look good. I ran the TADA_AutoClean function on a few WQP datasets (Upper Sioux, USGS MN, and Rogue River Volunteer group) and it ran as expected. It's hard for me to dig into the function code without having a better understanding of the various objects - but I think it makes sense. Something I noticed in the current UnitConversons.R is that L262-265 describe Speciation in USGS result units - this text may need to be revised to reflect the changes?

hillarymarler:
@ Thanks for the review and testing, @cefergus. I will take a look at the text you identified in L262-265.

PR #498, opened 2024-07-15
467-add-an-all-others-category-to-tada_fieldvaluespie-when-there-are-many-categories

wokenny13:
Added all other categories if number of categories exceed 12.


hillarymarler:
@wokenny13 - this looks great. Once the final check passes, I will merge it.

hillarymarler (approved):

PR #501, opened 2024-07-17
494 epatada continuous data flag error

hillarymarler:
New approach to flagging continuous data. Instead of relying on a loop, this version groups data (TADA.ComparableDataIdentifier, MonitoringLocation, various depth fields), arranges the date consecutively by ActivityStartDateTime and then calculates the time difference between each result and the result before and after it within each group.

There are some slight differences in the results between this function and the original. Additional results are identified as "discrete" in this version. I think this is correct as when I reviewed the data set, in each of these cases, the result labeled as "discrete" was the only one in its "group". I am not sure why they were identified as continuous in the previous function.

I have an example data set with results from the old function. It was too large to attach here, but I can send it if you would like to compare its results to results from the updated function.


R/ResultFlagsIndependent.R
```diff
@@ -153,11 +153,11 @@ TADA_FlagMethod <- function(.data, clean = TRUE, flaggedonly = FALSE) {
...
 #' @param time_difference Numeric argument defining the maximum time difference
-#' in hours between measurements taken on the same day. This is used to search for
-#' continuous time series data (i.e., if there are multiple measurements on the same
-#' day within the selected time_difference, then the row will be flagged as
-#' continuous). The default time window is 4 hours. The time_difference can be
-#' adjusted by the user.
+#' in hours between measurements of the same TADA.ComparableDataIdentifier taken at the same
+#' latitude, longitude, and depth. This is used to search for
+#' continuous time series data (i.e., if there are multiple measurements within the selected
+#' time_difference, then the row will be flagged as continuous). The default time window is 4 hours.
```
wokenny13, line 159: Would decimal values for the time windows work? I've tried an example with 0.25 and even 0.20 and it seems to have worked, but didn't play around too much with fractional times

wokenny13:
Ran through the examples with Data_Nutrients_UT and looked into using different time_difference rather than the default 4. Also ran through an own example dataset with just Florida, and the flagging of continuous data seems to have worked as expected. 

Run time was quick, and I think the options to flag or clean the dataset for continuous data is nice. I tested fractional time values, and it does run and seem to work, but perhaps fractional values are not of much interest.

wokenny13 (approved):
Ran through the examples with Data_Nutrients_UT and looked into using different time_difference rather than the default 4. Also ran through an own example dataset with just Florida, and the flagging of continuous data seems to have worked as expected.

Run time was quick, and I think the options to flag or clean the dataset for continuous data is nice. I tested fractional time values, and it does run and seem to work, but perhaps fractional values are not of much interest.

R/ResultFlagsIndependent.R
```diff
@@ -269,124 +254,100 @@ TADA_FlagContinuousData <- function(.data, clean = FALSE, flaggedonly = FALSE, t
...
-            ((noncont.data$TADA.ResultDepthHeightMeasure.MeasureValue == noncont.data$TADA.ResultDepthHeightMeasure.MeasureValue[i]) | (is.na(noncont.data$TADA.ResultDepthHeightMeasure.MeasureValue) & is.na(noncont.data$TADA.ResultDepthHeightMeasure.MeasureValue[i]))) &
-            ((noncont.data$TADA.ActivityTopDepthHeightMeasure.MeasureValue == noncont.data$TADA.ActivityTopDepthHeightMeasure.MeasureValue[i]) | (is.na(noncont.data$TADA.ActivityTopDepthHeightMeasure.MeasureValue) & is.na(noncont.data$TADA.ActivityTopDepthHeightMeasure.MeasureValue[i]))) &
-            ((noncont.data$TADA.ActivityBottomDepthHeightMeasure.MeasureValue == noncont.data$TADA.ActivityBottomDepthHeightMeasure.MeasureValue[i]) | (is.na(noncont.data$TADA.ActivityBottomDepthHeightMeasure.MeasureValue) & is.na(noncont.data$TADA.ActivityBottomDepthHeightMeasure.MeasureValue[i]))) &
-            ((noncont.data$ActivityRelativeDepthName == noncont.data$ActivityRelativeDepthName[i]) | (is.na(noncont.data$ActivityRelativeDepthName) & is.na(noncont.data$ActivityRelativeDepthName[i])))
-        )
+    info_match <- noncont.data %>%
+      dplyr::group_by(
+        TADA.LatitudeMeasure, TADA.LongitudeMeasure,
```
renaemyers, line 268: Question from Renae: Would there be any benefit to grouping by MonitoringLocationIdentifier instead of lat/long? Part of the reason I ask is, if the user had previously used the TADA_FlagNearbySites function to add column TADA.MonitoringLocationIdentifier, it could be used instead of MonitoringLocationIdentifier, which may be desirable?

Response from Hillary:  We had talked about using TADA.MonitoringLocationIdentifier, but after some discussion decided that grouping by lat/long and organization identifier would help ensure that samples identified as continuous are from the same probe/continuous monitoring effort instead of the unlikely (but possible) scenario of two organizations collecting their own discrete samples within the user-specified time window at nearby sites.

renaemyers (approved):
The filtering approach is much better than the loop that the function previously used. I had run times between a couple seconds using the Data_Nutrients_UT example data and 2.5 minutes using a large custom dataset of about 207,000 rows.

cefergus (approved):
I think the revised function looks good from what I could see. I ran the revised function on Fond du Lac data and a random TADA test data set. I looked for observations with same location, depth, comparable data identifier, and organization and it looked like the function is correctly flagging continuous vs discrete observations. Some observations labeled "Field Msr/Obs" and flagged as "Continuous" look like duplicated result values. But maybe a different function can flag those incidences.

hillarymarler:
Thanks for the reviews and comments, @renaemyers and @cefergus.

I know we need to revisit the find QA/QC and paired replicates functions again at some point in the future, so I will copy @cefergus comment to the appropriate paired replicates issue or create a new one, so that we can consider it in the context of future updates to be applied to QC or continuous flagging functions.

PR #507, opened 2024-08-06
504 consistent color pairs for outlinefill for figures

wokenny13:
Color pairings were done within the TADA_ColorPalette function using an optional argument. Reflected these changes in some of the .R Figures where it seemed appropriate.


hillarymarler:
@wokenny13 - I will review this PR tomorrow or Friday.

hillarymarler (approved):
@wokenny13 - this was a really nice and efficient solution. What would you think of adding the col_pair argument to the TADA_ViewColorPalette function? So that users/devs could easily preview the color pairs. Could be done as part of this PR or a future one.

wokenny13:
I think that's something that I would be able to add on to TADA_ViewColorPalette as part of this pr. I can work on completing that item either later today or early next week

hillarymarler:
Thanks - that sounds like a great plan. I will wait to merge until you have had a chance to add that feature.

wokenny13:
@hillarymarler I added the col_pair as a view for TADA_ViewColorPalette. I created the output as a matrix, but am open to suggestions on how it should be viewed if there are other thoughts.

For the example:

`#' TestViewPalette <- TADA_ViewColorPalette()
#' TestViewPalettePairing <- TADA_ViewColorPalette(col_pair = TRUE)`

I noticed that if you run it, and try to store it under the variable, for example 'TestViewPalette', that TestViewPalette returns NULL as the value. It still, initially, plots the view color palette function output, but if you try to call the variable it was stored under, 'TestViewPalette' only NULL is returned. Not sure if that's expected or not. 

I think the update that I made, plots both the original TADA_ViewColorPallete and the paired figures output, in separate plots. If you feel like it should only plot either the paired output or the original figure, depending on if col_pairs = TRUE/FALSE, I can fix that.

hillarymarler:
This looks great. I like that way you have the output set up. What would you think about swapping out tada palette #4 for 12 and 13 from the unpaired palette? To make sure they are all distinct enough.

image

R/Figures.R
```diff
@@ -1380,23 +1376,23 @@ TADA_GroupedScatterplot <- function(.data, group_col = "MonitoringLocationName",
         plot_bgcolor = "#e5ecf6",
         margin = mrg,
         legend = list(
+          title = list(text = paste0('<b>', group_col,'<b>'), x = 0.5, y= 100),
```
wokenny13, line 1379: @hillarymarler over here for legend title

hillarymarler: Thanks!

PR #511, opened 2024-08-16
Figures.r condensed code and tada view color palette to return a non null value

wokenny13:
Created a for loop for TADA_GroupedScatter to condense the code and make it more concise. Will work to edit other functions like TADA_DepthProfilePlot to see if this can be done there

TADA_ViewColorPalette() should also return a stored value in the example documentations now rather than return NULL or an empty list().

### DIFF
--- a/R/Figures.R
+++ b/R/Figures.R
@@ -1175,9 +1175,9 @@ TADA_TwoCharacteristicScatterplot <- function(.data, id_cols = "TADA.ComparableD
 #' data(Data_Nutrients_UT)
 #' # UT Nutrients results grouped by county
 #' # transform non-detect data
-#' df <- TADA_SimpleCensoredMethods(Data_Nutrients_UT)
+#' df2 <- TADA_SimpleCensoredMethods(Data_Nutrients_UT)
 #' # create scatterplots for selected counties
-#' UT_Nutrients_by_CountyCode <- TADA_GroupedScatterplot(df, group_col = "CountyCode", groups = c("057", "011", "003", "037"))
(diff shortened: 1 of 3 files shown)


wokenny13:
Waiting to see if checks go through successfully. Once it is successful, checking a few other examples to make sure the TADA_GroupedScatterplot is still functioning would be good to do.

hillarymarler (changes requested):
The changes to the grouped scatterplot are excellent- code is much more concise and it functions as intended.

For viewing the color palette, the updates work for the original palette, but I get this error when I run it for the pairs:
image

This happens when it run it with or without assigning it a name.

wokenny13:
> The changes to the grouped scatterplot are excellent- code is much more concise and it functions as intended.
> 
> For viewing the color palette, the updates work for the original palette, but I get this error when I run it for the pairs: image
> 
> This happens when it run it with or without assigning it a name.

Thanks for finding this! I will look into it to determine what may be the cause

hillarymarler:
I'm still seeing the same error when I try to assign a name with col_pair = TRUE.

wokenny13:
It seems like that error may show up depending on the screen margin of the monitor it is being viewed on. On my second monitor I don't get any error messages, but when I switch RStudio to the laptop monitor screen, I do run into an error. 

But when I do re-run it on the laptop monitor screen, I am able to view it, but it shows up looking like this:
image

Rather than looking like this on my 2nd monitor:
image

What does your plot look like when you run TADA_ViewColorPalette(col_pair = TRUE)?

I will try to see if I can find a solution by adjusting the figure margins to be compatible when viewed on any monitor

hillarymarler:
This is a tricky one!

On my laptop monitor I get this error:

image

I have two other monitors. One of them it works, on the other one, I get this error message:
image

hillarymarler (approved):

PR #514, opened 2024-08-23
513-occasional-test-failure-tada_findpotentialduplicates-does-not-grow-dataset

wokenny13:
NA logical value for TADA.MonitoringLocationIdentifier created in TADA_FindNearbySites() function it seems.

versus

"NA" character values for TADA.MonitoringLocationIdentifier in the dupsdat dataframe in this function.

Joining causes additional rows created in its return value due to mismatch of type.

Convert logical NA to character "NA".

If TADA_FindNearbySite does not return NA for TADA.MonitoringLocationIdentifier, it will return the respective TADA.MonitoringLocationIdentifier still.

### DIFF
--- a/R/ResultFlagsIndependent.R
+++ b/R/ResultFlagsIndependent.R
@@ -1281,7 +1281,9 @@ TADA_FindPotentialDuplicatesMultipleOrgs <- function(.data, dist_buffer = 100, o
       dplyr::mutate(
         TADA.MultipleOrgDuplicate = ifelse(is.na(TADA.MultipleOrgDuplicate), "N", TADA.MultipleOrgDuplicate),
         TADA.ResultSelectedMultipleOrgs = ifelse(is.na(TADA.ResultSelectedMultipleOrgs), "Y", TADA.ResultSelectedMultipleOrgs),
-        TADA.MultipleOrgDupGroupID = ifelse(is.na(TADA.MultipleOrgDupGroupID), "Not a duplicate", TADA.MultipleOrgDupGroupID)
+        TADA.MultipleOrgDupGroupID = ifelse(is.na(TADA.MultipleOrgDupGroupID), "Not a duplicate", TADA.MultipleOrgDupGroupID)) %>%
+      dplyr::mutate( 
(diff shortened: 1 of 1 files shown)


cristinamullin:
Good catch Kenny. We've run into similar issues before and applied a similar approach to resolve. We changed NA to "NA - Not Available" (character). 

Should we convert to "NA - Not Available" here as well for consistency with other TADA functions?

wokenny13:
Keeping any NA changes to "NA - Not Available" for consistency sounds like it would be good.

In general, does storing a value as a logical/missing NA versus a character NA make any difference in speed/performance/memory?

**Best spot to handle this change?**

I think dupsdata convert the logical NA for TADA.MonitoringLocationIdentifier to a character in this chunk of the code in TADA_FindPotentialDuplicatesMultipleOrgs. In this case, since a full join is used at the end of the code, I believe it would require then changing "NA" found in dupsdata to "NA - Not Available"

`    dupsdat <- dupsdat %>%
      dplyr::rename(SingleNearbyGroup = TADA.MonitoringLocationIdentifier) %>%
      dplyr::mutate(
        TADA.MonitoringLocationIdentifier = paste(SingleNearbyGroup, sep = ","),
        TADA.ResultSelectedMultipleOrgs = ifelse(ResultIdentifier %in% duppicks$ResultIdentifier, "Y", "N")
      ) %>%
      dplyr::select(-SingleNearbyGroup)
`

Or would we want to make edits to NA value in lines 841 under TADA_FindNearbySites() where 

`if (!"TADA.MonitoringLocationIdentifier" %in% colnames(.data)) {
    .data$TADA.MonitoringLocationIdentifier <- NA
  }`

hillarymarler:
@wokenny13 - is this ready for review?

wokenny13:
@hillarymarler 

> Good catch Kenny. We've run into similar issues before and applied a similar approach to resolve. We changed NA to "NA - Not Available" (character).
> 
> Should we convert to "NA - Not Available" here as well for consistency with other TADA functions?

I will work on converting it to "NA - Not Available" for consistency. I will let you know when I get this finished and for it to be reviewed then

hillarymarler:
Sounds great - thank you!

wokenny13:
@hillarymarler this is ready for review

hillarymarler (approved):

PR #518, opened 2024-08-27
512 tada convertresultunits returns error when argument ref = wqx

hillarymarler:
Fixed bug when ref = "wqx". Removed unnecessary column from unit.ref when ref = "wqx" to fix joining error.


wokenny13:
The wqx = ref is working correctly now with this update

wokenny13 (approved):

PR #523, opened 2024-09-10
482 create tadamonitoringlocationidentifier in tada autoclean

hillarymarler:
I am still working on updating functions which rely on either "MonitoringLocationIdentifier" or "MonitoringLocationName" to use the TADA columns instead. However, the updates to TADA_FindNearbySites are working on test data sets, so any thoughts on how to improve those would be helpful. I have not incorporated using the AlternateMonitoringLocationID field as that isn't currently a field we include in the TADA df.

Should TADA_OverviewMap switch to using TADA.MonitoringLocationIdentifier so that grouped sites show up with only one marker on the map? Or should I add a param so that users can select whether the original or TADA monitoring location are mapped?


hillarymarler:
This is ready for review. If you spot any additional places where I have not made the MonitoringLocationIdentifier -> TADA.MonitoringLocationIdentitifer switch.

hillarymarler:
Closing this PR. Will resubmit when I have a working draft involving the geospatial components we've discussed in Team meetings.

PR #535, opened 2024-10-22
534 dataset growing mt example

hillarymarler:
Update to TADA_ConvertResultUnits to prevent the dataset from growing when converting result units or using autoclean. 

The problem as many-to-many relationships between detection limit units and the conversion reference. Adding "TADA.DetectionQuantitationLimitMeasure.MeasureUnitCode" as another joining column prevents many-to-manys at this join.


wokenny13 (approved):
I ran through several dataset tests, and everything looks like it is running as it should now.

hillarymarler:
Thanks @wokenny13 - I am going to go ahead and merge this since it was a user reported bug and they're eager to get back to work! We can discuss if additional updates are needed with the rest of the team later this week.

PR #537, opened 2024-10-29
Hrm test

hillarymarler:
This is an option to automate checking URLs in the package as a testthat test. I corrected any URLs that had been moved/redirected. There were a couple of URLs I excluded from the test:

 "https://www.itecmembers.org/attains/" - something about its server configuration is not allowing it to send the correct HTTP status code. I tried a couple of different methods and could not get any response, though the site loads fine if you paste the link in a browser.

And the http and https links to cran as they always come back with a redirect message (I think due to automatic redirect to a mirror). I can modify the test to accept/not flag redirects (300s) if that is preferable for our use case. Currently, the test looks for 200 response and anything other than that will fail.

If the test were to fail, the easiest way to find the failing URL would be to open the package and manually run lines 3-53 in the test-URLChecker.R file. The "df_false" data frame will contain any failing urls and their respective http response codes.


hillarymarler:
Unfortunately, the new test failed here (though not while I was testing it). I will work on getting it fixed so it can be reviewed.

cristinamullin:
Great work Hillary! Redirects (300s) should be fine for our our use case - do you want to modify the code to accept those now or wait to see if it becomes burdensome to edit the redirects in the future? 

I think it'd be helpful to add directions (below) on how to fix the broken urls to Maintenance.R. The code / any notes in Maintenance.R will need to be commented out (like the styler code in there now), but I think it's helpful to have all this in one spot.

hillarymarler:
Sure - I can modify so that redirects are accepted too. This would mean I can remove CRAN from the exclusion list and exclude only the ITEC site from the test. I can also add instructions on how to identify and fix broken links if this test fails in Maintenance.R.

cristinamullin (approved):

PR #540, opened 2024-11-07
397 range check bug

hillarymarler:
Updates both TADA threshold functions now that validation table has been updated to ensure that there is only one min/max for each characteristic/source/value.unit combination.


cristinamullin:
@hillarymarler this all look great & is ready to merge once the checks complete

PR #545, opened 2024-11-14
Q ctable test

hillarymarler:
Validation table test. Instructions for how to identify causes of test failure are included in Maintenance.R. Currently, the Maintenance.R file instructions/code are commented out. They include creating and saving a csv to share with the wqx team. I created a testresults folder in "testthat" to save the results. But I could also set it up to save in the downloads folder if that is preferable.


cristinamullin:
I think I'd prefer this to be saved to the downloads folder so that it does not accidentally get included in the package (cluttering/taking up space). Alternatively the folder itself and file could be added to .gitignore so that doesn't happen. Either works

hillarymarler:
I'll get it switched to the downloads folder - that makes the most sense overall, the more I think it about it.

cristinamullin (approved):

PR #549, opened 2024-11-29
Update continuous data flagging

hillarymarler:
Added filter to exclude quality control samples from continuous data search. I have not been able to recreate the issue with all NAs in continuous flag column by using examples from previous test failures or random data sets.

### DIFF
--- a/R/ResultFlagsIndependent.R
+++ b/R/ResultFlagsIndependent.R
@@ -260,13 +260,20 @@ TADA_FlagContinuousData <- function(.data, clean = FALSE, flaggedonly = FALSE, t
 
   # everything not YET in cont dataframe
   noncont.data <- subset(.data, !.data$ResultIdentifier %in% cont.data$ResultIdentifier)
+  
+  # import WQX Activity Type Ref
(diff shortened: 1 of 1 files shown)


R/ResultFlagsIndependent.R
```diff
@@ -265,6 +265,8 @@ TADA_FlagContinuousData <- function(.data, clean = FALSE, flaggedonly = FALSE, t
 
   if (length(noncont.data) >= 1) {
     info_match <- noncont.data %>%
+      # remove quality control samples
+      dplyr::filter(!grepl("Quality Control", ActivityTypeCode)) %>%
```
cristinamullin, line 269: There are a few ActivityTypeCode's that do not include "Quality Control" that are also QC samples. Could you reference the internal TADA package WQXActivityTypeRef here instead and keep only the results assigned by TADA as "Non_QC"?

hillarymarler: Yes - that is a better, more specific solution. I can make that change.

hillarymarler:
Updated to filter using TADA.ActivityCode.Flag

cristinamullin (approved):

PR #553, opened 2024-12-12
Tada shiny flag tab bug fix

cristinamullin:


hillarymarler (changes requested):
Your updates look good and make sense to me. Do you think we can delete the couple of sections you commented out (reading in Activity Type Ref, joining it)?

PR #560, opened 2024-12-26
482 create tadamonitoringlocationidentifier in tada autoclean

hillarymarler:
Updates to TADA_FindNearby - submitting PR as I am having trouble running check() locally. I anticipate I will need to make some additional changes before it is ready for review.


wokenny13:
I will review later today once all the checks all good

hillarymarler:
@wokenny13 - it seems like the only lingering check issue is due to the process timing out, so I think you'd be ok to go ahead and review the function updates anytime you are ready. I think fixing the timeout issue is going to involve updates of some kind to vignettes/example data like we've previously discussed.

wokenny13:
I took a quick look through, this looks good to merge with #568. 

Maybe we can discuss in the future whether TADA_FlaggedSitesMap could use TADA.MonitoringLocationIdentifier or just MonitoringLocationIdentifier as an argument input. And likewise, on if there are use cases when any other functions that runs TADA_FindNearbySites to create TADA.MonitoringLocationIdentifier are needed or could be used as an argument input? The runtime could be a bit long from what I've seen? 

TADA_FindNearbySites(Data_Nutrients_UT) seems like a pretty long runtime in the example. Can a smaller example dataframe be used? Or filter this data frame down to reduce its runtime?

hillarymarler:
The major disadvantage to the adjacency matrix approach now used in TADA_FindNearbySites is that it takes longer to find groups of related sites than just identifying sites within a buffer distance of each other. The main improvement that is represents is the function is no  longer assigning the same site to multiple groups. So, we may need to discuss that tradeoff more as a group. If we can't get to that before the conference, I don't think that holding off on this PR would have a negative impact on any of the planned examples.

wokenny13:
I think the updates are nice, to avoid having same sites in multiple groups. However, it may be good to have a further discussion on this in one of our group meetings to think through if the longer run time is a fine tradeoff.

cristinamullin:
Changes in this branch will be merged in as part of this PR: https://github.com/USEPA/EPATADA/pull/575

PR #562, opened 2024-12-30
Au ml crosswalk

hillarymarler:
Draft functions related to monitoring location/Assessment Unit crosswalks:

1) TADA_GetAssessmentUnitCrosswalk - using ATTAINS webservices, creates a crosswalk between Assessment Units and all Monitoring Locations recorded for them in ATTAINS

2) TADA_GetMonLocByOrgId - create a df of MonitoringLocationIdentifier and OrganizationIdentifier to help w/ finalizing user created crosswalk (not sure if this is necessary? I was using it to help create test crosswalks)

3) TADA_GetProviderRef - using WQX web services to create a ref file for the ProviderName for each Organizations (used later on to build the URLs for each MonitoringLocation to link to the WQP site pages)

4) TADA_UpdateProviderRef - update the provider ref csv file saved in the package

5) TADA_UpdateMonitoringLocationsInATTAINS - use a user-supplied Monitoring Location crosswalk w/ the ATTAINS created crosswalk to create a df to update Monitoring Location info in ATTAINS. I have not yet set up this function to save as a csv. I am thinking that as we will likely be setting up a number of similar functions in the future a TADA_CreateCSV function might be useful to create for the simple batch import csvs (in constrast to the complex mod 3 Excel files, which will need the more complicated export and formatting code built into each of those functions)


hillarymarler:
@wokenny13 - I added to TADA_GetAssessmentUnitCrosswalk to add the org_id prefix to MonitoringLocationIdentifiers where appropriate.

I need to do an additional update to TADA_UpdateMonitoringLocationsinATTAINS related to when crosswalk = NULL.

R/CrosswalksFromATTAINS.R
```diff
@@ -0,0 +1,435 @@
...
+#' @export
+#'
+#' @examples
+#' \dontrun{
+#' # Alaska example
+#' AK_crosswalk <- TADA_GetAssessmentUnitCrosswalk(org_id = "AKDECWQ")
+#' }
+
+#'
+TADA_GetAssessmentUnitCrosswalk <- function(org_id = NULL) {
```
cristinamullin, line 26: What do you think about possibly using this as the function name instead? I think it would be helpful to clarify in the name that this is coming from ATTAINS in the name.

TADA_GetATTAINSAUSiteCrosswalk

R/CrosswalksFromATTAINS.R
```diff
@@ -76,6 +87,7 @@ TADA_GetAssessmentUnitCrosswalk <- function(org_id = NULL) {
         "No MonitoringLocationIdentifiers were recorded in ATTAINS for ",
         org_id, " Assessment Units."
       ))
+      return(au.crosswalk)
```
cristinamullin, line 90: I updated this so it returns a blank df now if there is no crosswalk for the org. This is consistent with other TADA functions. Any issues with using that approach here? @hillarymarler

R/CrosswalksFromATTAINS.R
hillarymarler: I like that name better - I struggled with coming up with a descriptive, concise function name here and this covers it all!

R/CrosswalksFromATTAINS.R
hillarymarler: I don't think that will cause any problems- initially I was thinking a blank crosswalk might be useful if the user didn't have any MLs in ATTAINS, but the mod 3 functions facilitate creating a crosswalk much more efficiently than using any blank option generated here.

R/CrosswalksFromATTAINS.R
cristinamullin: Great! I'll add this to my next commit.

cristinamullin (approved):

DESCRIPTION
```diff
@@ -72,7 +72,8 @@ Imports:
     nhdplusTools,
     arcgislayers,
     httr,
-    jsonlite
+    jsonlite,
+    rATTAINS
```
cristinamullin, line 76: After our meeting last Friday, I think we should remove rATTAINS as a dependency. We can copy the code for functions leveraged and include them as functions in EPATADA.

R/CrosswalksFromATTAINS.R
```diff
@@ -0,0 +1,418 @@
...
+#' This function creates a crosswalk of OrganizationIdentifier and OrganizationFormalName with
+#' ProviderName.
+#'
+#' @return A crosswalk of monitoring locations and organization identifiers.
+#'
+#' @export
+#'
+#' @examples
+#' \dontrun{
+#' provider.ref <- TADA_GetProviderRef()
```
cristinamullin, line 107: @hillarymarler This function doesn't seem to be working yet. Should it be included in this PR?

DESCRIPTION
hillarymarler: That sounds good to me - I can then work on updating to use EQ web services as time permits. Some of the functions that use rATTAINS may also be served well be  using the EQ national extracts. I started looking at saving those as .RData files earlier today.

R/CrosswalksFromATTAINS.R
hillarymarler: I can take a look at these again now/tomorrow, but if you want to merge in the others, I can fix these as a separate PR. Whatever works best for you!

wokenny13: > I can take a look at these again now/tomorrow, but if you want to merge in the others, I can fix these as a separate PR. Whatever works best for you!

If you insert this line:

# Used to store cached WQX Provider Reference Table
WQXProviderRef_Cached  <- NULL

before the function

 TADA_GetProviderRef <- function() {...

It seems like it would run then. But unsure of what the best practice is for where to store a cache ref table

hillarymarler: Thank you @wokenny13 - I probably left that out when I was borrowing the formatting for the update functions from the WQX ones.

R/Utilities.R
```diff
@@ -1699,3 +1704,64 @@ TADA_CharStringRemoveNA <- function(char_string) {
...
+#'
+#'
+#' @return A crosswalk dataframe of monitoring location identifiers and organization identifiers.
+#' 
+#' @export
+#'
+#' @examples
+#' data(Data_6Tribes_5y)
+#' orgsite_crosswalk_originalwqp = TADA_GetMonLocByOrgId(Data_6Tribes_5y, id = "wqp")
+#' orgsite_crosswalk_tada = TADA_GetMonLocByOrgId(Data_6Tribes_5y, id = "tada")
```
wokenny13, line 1732: TADA.MonitoringLocationIdentifier does not exist yet, is this correct? So the check does not pass through with TADA_GetMonLocByOrgId(Data_6Tribes_5y, id = "tada")

hillarymarler: Yes - it won't pass yet. I can switch it to "MonitoringLocationIdentifier" for now if that makes the most sense and then update once the TADA.MonitoringLocationIdentifier is merged. Initially, I expected that one to be finished first (but then it ended up being a lot more complicated than I expected)

cristinamullin: Correct. I just updated the example data to use Data_6Tribes_5y_Harmonized instead. Now, you can return either. One returns 227 obs and the other 203. This makes sense because some of the sites are grouped as "nearby" in the example dataset. @hillarymarler can you double check that this the desired/useful output for concatenated sites? Do you think we should add something about this to the function documentation? 

```r
data(Data_6Tribes_5y_Harmonized)
orgsite_crosswalk_originalwqp = TADA_GetMonLocByOrgId(Data_6Tribes_5y_Harmonized, id = "wqp")
orgsite_crosswalk_tada = TADA_GetMonLocByOrgId(Data_6Tribes_5y_Harmonized, id = "tada")
```

hillarymarler: Sure - I can double check and improve documentation. And I will do some more thinking about how the grouped sites should be handled. That might be a good thing to chat about in our next working session.

R/CrosswalksFromATTAINS.R
```diff
@@ -0,0 +1,423 @@
...
+#' }
+#'
+TADA_GetProviderRef <- function() {
```
cristinamullin, line 111: For a function name, how about TADA_GetWQPOrgProviderCrosswalk ? Any other suggestions? Okay to move this to WQXRefTables.R since it doesn't pull from ATTAINS?

R/CrosswalksFromATTAINS.R
```diff
@@ -0,0 +1,423 @@
...
+#'
+#' @export
+#'
+#' @examples
+#' \dontrun{
+#' # Alaska example
+#' AK_crosswalk <- TADA_GetATTAINSAUSiteCrosswalk(org_id = "AKDECWQ")
+#' }
+#'
+TADA_UpdateMonitoringLocationsInATTAINS <- function(org_id = NULL,
```
cristinamullin, line 196: This isn't working quite right. The example was for a different function so I just updated that, but it still doesn't work. See:
```r
AK_batchAUupload <- TADA_UpdateMonitoringLocationsInATTAINS(org_id = "AKDECWQ", crosswalk = NULL, attains_replace = FALSE, data_links = "update")
Error in TADA_UpdateMonitoringLocationsInATTAINS(org_id = "AKDECWQ", crosswalk = NULL,  : 
  object 'update.crosswalk' not found
```

R/CrosswalksFromATTAINS.R
```diff
@@ -0,0 +1,423 @@
...
+        org_id, " Assessment Units."
+      ))
+      return(au.crosswalk)
+    }
+  }
+}
+
+WQXProviderRef_Cached <- NULL
+
+#' Get Provider Ref for All Organizations (IN ACTIVE DEVELOPMENT)
```
cristinamullin: Does this one still need (IN ACTIVE DEVELOPMENT)?

R/CrosswalksFromATTAINS.R
```diff
@@ -0,0 +1,425 @@
...
+#' possible without updating other information in ATTAINS.
+#'
+#' @return The csv batch upload files for ATTAINS to update Monitoring Locations.
+#'
+#' @export
+#'
+#' @examples
+#' \dontrun{
+#' # Alaska example
+#' AK_batchAUupload <- TADA_UpdateMonitoringLocationsInATTAINS(org_id = "AKDECWQ", crosswalk = NULL, attains_replace = FALSE, data_links = "update")
```
cristinamullin, line 195: Can you add a few more examples? Including one with a user provided crosswalk (maybe create a simple df for the example with 5 rows of dummy data for the AU/site crosswalk). I am curious what that looks like.  

```r
AK_batchAUupload <- TADA_UpdateMonitoringLocationsInATTAINS(org_id = "AKDECWQ", attains_replace = TRUE, data_links = "update")
Error in TADA_UpdateMonitoringLocationsInATTAINS(org_id = "AKDECWQ", attains_replace = TRUE,  : 
  TADA_UpdateMonitoringLocationsInATTAINS: in order to replace MonitoringLocations stored in ATTAINS (with attains_replace = TRUE), user must provide a MonitoringLocation/AssessmentUnitcrosswalk.
```

R/CrosswalksFromATTAINS.R
```diff
@@ -0,0 +1,425 @@
...
+#'
+#' @param data_links Character argument. When data_links is equal to "update" 
+#' or "replace", the function will build the URL for the Water Quality Portal 
+#' Data Site page for each Monitoring Location in the df. It will examine the 
+#' response codes of the URLs and only retain those with a 200 response, which 
+#' indicates they are valid urls. When data_links = "update", the url will be 
+#' added to any existing text in the MS_DATA_LINK_TEXT column. When 
+#' data_links = "replace", the url will replace any existing text in the
+#' MS_DATA_LINK_TEXT column. When data_links = "none", no URLs will be created
+#' or added to the df. Default is data_links = "update".
```
cristinamullin, line 175: I changed "update" to "add" for this function input because update was confusing to me and felt too similar of a term to the other "replace" option.

cristinamullin: I also change data_links to wqp_data_links for clarity

R/CrosswalksFromATTAINS.R
```diff
@@ -0,0 +1,425 @@
...
+#'
+#' @param attains_replace Character argument. When attains_replace = FALSE, all 
+#' Monitoring Location Identifiers associated with an Assessment Unit in ATTAINS
+#' will be retained even if they are not included in the user supplied crosswalk.
```
cristinamullin, line 180: This part of the documentation is still a little confusing to me. 

"When attains_replace = FALSE, all  Monitoring Location Identifiers associated with an Assessment Unit in ATTAINS will be retained even if they are not included in the user supplied crosswalk."

Does this mean that new ones from the user supplied crosswalk will be appended (added) to the existing ones in ATTAINS so the output includes both?

R/CrosswalksFromATTAINS.R
```diff
@@ -1,13 +1,13 @@
...
-#' Assessment Units in ATTAINS, this function can be used to create a crosswalk of known
-#' MonitoringLocationIdentifiers and Assessment Units. All tribal nations record this 
-#' crosswalk in ATTAINS but only a few states. If a state has not supplied this 
-#' crosswalk to ATTAINS, the function will return a blank dataframe.
+#' through EPA ATTAINS may also submit a crosswalk of WQP MonitoringLocationIdentifiers associated 
+#' with their Assessment Units to ATTAINS. If the organization has recorded 
+#' MonitoringLocationIdentifiers associated with their Assessment Units in ATTAINS, this function 
+#' can be used to create a crosswalk of known MonitoringLocationIdentifiers and Assessment Units. 
+#' All tribal nations record this crosswalk in ATTAINS but only a few states. If a state has not 
+#' supplied this crosswalk to ATTAINS, the function not return a data frame.
```
cristinamullin, line 10: the function **will** not return a data frame

DESCRIPTION
hillarymarler: I will copy code for rATTAINS functions in a separate PR.

R/CrosswalksFromATTAINS.R
hillarymarler: Updated function name to TADA_GetWQPOrgProviderRef and moved to WQPWQXRefTables.R (previously WQXRefTables.R).

R/CrosswalksFromATTAINS.R
hillarymarler: Thanks - this function is working now.

R/CrosswalksFromATTAINS.R
hillarymarler: I removed "(IN ACTIVE DEVELOPMENT)" here

R/CrosswalksFromATTAINS.R
hillarymarler: Updated to 
"#' @param attains_replace Character argument. When attains_replace = FALSE, all 
#' Monitoring Location Identifiers in the user supplied crosswalk will be appended to the
#' existing ATTAINS crosswalk. When attains_replace = TRUE, Monitoring Location Identifiers 
#' will only be retained if they are in the user supplied crosswalk. Default equals FALSE."

R/CrosswalksFromATTAINS.R
hillarymarler: Corrected

R/Utilities.R
hillarymarler: Added some more detail to documentation. I had written this with the idea using it as a helper function to assist in matching more MLs with valid URLs for the WQP site pages, though I have not incorporated it for that purpose yet. I think as a group we should have a quick conversation about how I should do this (or if I should). So I'm ok with commenting this out for now, not exporting it, or moving it to a separate branch (to be incorporated later) if we don't want to add it/add it as an exported function.

R/CrosswalksFromATTAINS.R
hillarymarler: This example has been created and code updated to allow for users to include data links in their crosswalk.

hillarymarler:
This is ready for review of changes I made after it passes checks! Thanks!

hillarymarler:
@cristinamullin are there any additional changes you'd like me to make?

cristinamullin:
> @cristinamullin are there any additional changes you'd like me to make?

No, this looks great! I'll merge it in now

PR #574, opened 2025-02-27
update citation

cristinamullin:


cristinamullin:
After the new USGS dataRetrieval related issues are resolved, these edits will be pushed through from this branch/PR: https://github.com/USEPA/EPATADA/pull/575

PR #577, opened 2025-03-11
Update ATTAINSCrosswalks.R

hillarymarler:
Wendy pointed out that DOMAINS link for ATTAINS needed to be updated after a recent update to it from the ATTAINS team. I will look into options for making that link a stable URL on drupal as that would benefit both ATTAINS and TADA, but this updated link is correct for now.


cristinamullin:
Changes in this branch will be merged in with the "prerelease" branch

Issue #427, opened 2024-03-19
Add bounding box query option to TADA DR


cristinamullin:
**Is your feature request related to a problem? Please describe.**

The WQP allows users to query by a bounding box. Let's add this option to TADA_DataRetrieval.

Issue #361, opened 2023-11-14
Advanced geospatial WQP data retrieval via user supplied shapefile or ESRI WFS


cristinamullin:
**Is your feature request related to a problem? Please describe.**

Summary: Generate query via a user supplied shapefile input, csv with monitoring location IDs (dependent on addressing big DR and max URL issues first: https://github.com/USEPA/TADA/issues/363), or by using an ESRI WFS such as the EPA tribal boundary service. Potentially leverage DR/NLDI/nhdplus tools… or EPA WATERS which has a copy of the WQP sites to get the site IDs within the shapefile boundary and then query the WQP by those specific site IDs.

More info: 
Advanced WQP data retrieval: generate query via a user supplied shapefile input, csv with monitoring location IDs, or by using an ESRI WFS such as the EPA tribal boundary service. Potentially leverage EPA WATERS which has a copy of the WQP sites to get the site IDs within the shapefile boundary and then query the WQP by those specific site IDs. 

Example request from MT DEQ: 
For specific projects, we often have an area of interest where we want all of the sites from that area. Can the package or shiny app select all the data in a particular area by uploading a shapefile? Or alternatively, could you load up activityid’s for your area of interest, have it load the data for all of those activityid’s from the wq portal, and complete the shiny steps for those? 

Just thinking of a particular project of a watershed above a mining area, where I have to do QA/QC, and I would like to run the data for those activityid’s through the shiny steps, but they don’t conform to a particular HUC or assessment unit.  I wanted to see what type of result I get with the shiny app (compared to what I ended up doing in excel).  

**Describe the solution you'd like**

Supply a shapefile as an input for the data retrieval function and get back all sites within the selected areas. 

New features should include all of the following work:

-   [ ] Create the function/code.

-   [ ] Document all code using comments to describe what is does.

-   [ ] Create tests in tests folder.

-   [ ] Create help file using roxygen2 above code.

-   [ ] Create working examples in help file (via roxygen2).

-   [ ] Add to appropriate vignette (or create new one).

jbousquin:
This (query by bounding box extent) is typically how I'm using harmonize-wq. wrangle.get_bounding_box() just pulls the extent off the shp in dataretrievel-python package's wqp.get_results() expected bBox order/format: 'xmin,ymin,xmax,ymax'

[...]

Issue #345, opened 2023-10-05
Query all data within tribal land boundary


cristinamullin:
**Is your feature request related to a problem? Please describe.**

Tribes need to be able to download all data within their tribal boundary. Currently, this is not possible with the available WQP filters. 

**Describe the solution you'd like**

Use EPA tribal land boundary service to get WQP station IDs within the boundary, then query the WQP to get all sites within the boundary. A new function input could be added to TADA_DataRetrieval in DataDiscoveryRetrieval.R to facilitate this (e.g. tribalBoundary = "FondduLac"). Meaning, the user selects the tribal name, and the resulting df would include only WQP sites that are within their tribal land boundary.

**Describe alternatives you've considered**

Find another way to do this, or ask the WQP team if we can do it in FY24 (unlikely). 

**Additional context**

The HMW is interested in facilitating this as well. 
[9/28 1:32 PM] Brad Cooper
WATERS/RAD (a little behind "processing schedule") has a copy of all WQP locations, we could use that EPA WQP service and do the query to get a list of station ids and then use those station ids to formulate the WQP data download web service call.

https://www.epa.gov/geospatial/guidance-using-tribal-boundaries-areas-and-names-resources 

**Reminders for TADA contributors addressing this issue**

New features should include all of the following work:

-   [ ] Create the function/code.

-   [ ] Document all code using comments to describe what is does.

-   [ ] Create tests in tests folder.

-   [ ] Create help file using roxygen2 above code.

-   [ ] Create working examples in help file (via roxygen2).

-   [ ] Add to appropriate vignette (or create new one).

Issue #159, opened 2022-11-08
Data retrieval size and progress bar


edhinman:
Wondering if it might be helpful to have a size cap on the data retrieval function or a warning that tells people when they're asking for a dataset that might cause the system to choke? Likely depends upon the machine's memory and such, but I have submitted many queries that cause my (very feeble) computer to freeze up because it can't handle the pull. This would require pulling stats from WQP before data download, so may not be an easy fix. You could have an example ideal dataset "size" as well?

cristinamullin:
@ldecicco-USGS What are your thoughts about this issue? Does dataRetrieval provide any recommendations for users on this?

ldecicco-USGS:
It's not always the size of the data - sometimes the complexity of the query can really slow things down and cause a timeout. It's hard to provide a specific recommendation because the underlying databases *do* get optimized over time (so the things we use to recommend are now possibly obsolete).

Ideally, you can use the summary service from WQP to roughly figure out how much data you are going to get, then you can chunk the data into smaller queries. If you pick a "chunk" size that your computer can't handle, lower the chunk size.

The 2 articles that talk about big data pulls do that in different ways:

Scripting:
https://rconnect.usgs.gov/dataRetrieval/articles/wqp_large_pull_script.html

Data-pipeline:
https://rconnect.usgs.gov/dataRetrieval/articles/wqp_large_pull_targets.html

cristinamullin:
Thanks Laura! That was my understanding as well. This is definitely a tough problem. 

I took the scripting approach when developing the [TADABigdataRetrieval](https://usepa.github.io/TADA/reference/TADABigdataRetrieval.html) function & actually referenced that USGS Scripting blog (https://rconnect.usgs.gov/dataRetrieval/articles/wqp_large_pull_script.html) heavily when developing it. 

Elise, I wonder if TADABigdataRetrieval might work for your big queries as is, or if it would need to be updated? Can you share an example query that usually times out for me to test? 

Note: Currently I only included startDate, endDate, characteristicName, and siteType as the query input options for this function. But I may be able to add more if needed. My initial thought here was to limit the query filters upon retrieval, and then filter down after if needed. We are also somewhat limited with the scripting approach to using filters that are available through the summary service (but I can look into the logistics of all that later).

edhinman:
This is the type of query we might run to obtain all data to consider for the IR over the period of record:

[...]

PR #578, opened 2025-03-13
Mini Release

cristinamullin:
- Updates the DOMAINS link for ATTAINS: Hillary is looking into options for making that link a stable URL on drupal so we can leverage that in the future as that would benefit both ATTAINS and TADA, but this updated link is correct for now.
- Fixes a bug in the harmonization reference file: now assumes "NITRATE + NITRITE" speciation is "as N" for all known combinations if left blank.
- Incorporates USGS dataRetrieval updates: we are now leveraging their "develop" branch for EPATADA instead of CRAN.
- Updates the citation: now includes new collaborators.
- Updates TADA_FindNearbySites: The major disadvantage to the adjacency matrix approach now used in TADA_FindNearbySites is that it takes longer to find groups of related sites than just identifying sites within a buffer distance of each other. The main improvement that it represents is the function is no longer assigning the same site to multiple groups. So, we may need to discuss that tradeoff more as a group.
- Creates tadamonitoringlocationidentifier in tada autoclean
- Includes significant updates to TADA_dataRetrieval

1. adds sf option, 
2. adds tribal options
3. updates big data options

Overview
- Addition of {sf} methods to allow users to query WQP data using {sf} objects
- Addition of options to allow tribal lands to be more directly queried using TADA_DataRetrieval
- New function, TADA_TribalOptions, to assist users with identifying and querying tribal lands
- Folding the processes in TADA_BigDataRetrieval into TADA_DataRetrieval and removing TADA_BigDataRetrieval to avoid confusion
- Adding progress bar to large data pulls, user prompt to confirm download, silencing {dataRetrieval} messages + error handling for HTTP errors, vignette update

Additional info

- {sf} methods use aoi_sf arg and largely begin [here](https://github.com/rossyndicate/EPATADA/blob/62973351d0bfaae2d51ff1d6273f1e3b304821b6/R/DataDiscoveryRetrieval.R#L280). First checks what data are available for the [bbox](https://github.com/rossyndicate/EPATADA/blob/62973351d0bfaae2d51ff1d6273f1e3b304821b6/R/DataDiscoveryRetrieval.R#L442) of the {sf} object provided, then uses only MonitoringLocationIdentifiers [inside](https://github.com/rossyndicate/EPATADA/blob/62973351d0bfaae2d51ff1d6273f1e3b304821b6/R/DataDiscoveryRetrieval.R#L486) the {sf} object when running the full query
- Tribal land queries use tribal_area_type and tribe_name_parcel args and are handled [alongside](https://github.com/rossyndicate/EPATADA/blob/62973351d0bfaae2d51ff1d6273f1e3b304821b6/R/DataDiscoveryRetrieval.R#L280) {sf} because they use [this EPA spatial data](https://geopub.epa.gov/arcgis/rest/services/EMEF/Tribal/MapServer). Both tribal_area_type and tribe_name_parcel are required. {sf} and tribal args can't be used at the same time (error), and if geographic info like statecode are provided in addition to either {sf} or tribal args then a [warning is returned](https://github.com/rossyndicate/EPATADA/blob/62973351d0bfaae2d51ff1d6273f1e3b304821b6/R/DataDiscoveryRetrieval.R#L230-L262)
- tribal_area_type refers to one of the EMEF/Tribal MapServer layers. tribe_name_parcel refers to either TRIBE_NAME or PARCEL_NO entries from that layer. The [TADA_TribalOptions](https://github.com/rossyndicate/EPATADA/blob/62973351d0bfaae2d51ff1d6273f1e3b304821b6/R/DataDiscoveryRetrieval.R#L953) function is included to help users see TRIBE_NAME/PARCEL_NO options available to them and check punctuation, etc.
- [TADA_BigDataHelper](https://github.com/rossyndicate/EPATADA/blob/62973351d0bfaae2d51ff1d6273f1e3b304821b6/R/DataDiscoveryRetrieval.R#L1106) is now used to handle "big" data requests [within](https://github.com/rossyndicate/EPATADA/blob/62973351d0bfaae2d51ff1d6273f1e3b304821b6/R/DataDiscoveryRetrieval.R#L517-L536) TADA_DataRetrieval. By default this is triggered with maxrecs = 250000 & maxsites = 300.
- Two ([1](https://github.com/rossyndicate/EPATADA/blob/62973351d0bfaae2d51ff1d6273f1e3b304821b6/R/DataDiscoveryRetrieval.R#L1157), [2](https://github.com/rossyndicate/EPATADA/blob/62973351d0bfaae2d51ff1d6273f1e3b304821b6/R/DataDiscoveryRetrieval.R#L1201)) progress bars are included inside TADA_BigDataHelper
- The [ask_user](https://github.com/rossyndicate/EPATADA/blob/62973351d0bfaae2d51ff1d6273f1e3b304821b6/R/DataDiscoveryRetrieval.R#L1326) function is used to [confirm](https://github.com/rossyndicate/EPATADA/blob/62973351d0bfaae2d51ff1d6273f1e3b304821b6/R/DataDiscoveryRetrieval.R#L500) that the user wants to download the dataset after the number of records is determined
- In general the messages from {dataRetrieval} are now silenced because they were returning a lot of information that was hiding (what we considered) more useful information from TADA_DataRetrieval. But we've made sure to include [checks for HTTP errors](https://github.com/rossyndicate/EPATADA/blob/62973351d0bfaae2d51ff1d6273f1e3b304821b6/R/DataDiscoveryRetrieval.R#L439-L457), which will then be communicated back to the user
- Additional info now in vignette 1 to explain the new {sf}, tribal, and big data functionality

A few notes:
- I left NULL as the default for the aoi_sf argument instead of "null" because the character version didn't work properly
- I had hoped to work on issues related to character length limits in queries, as discussed with Cristina, but ran out of time
- From my tests it didn't seem like the way that data are indexed by calendar date affected query speed
- Please let me know if I can provide any other info on any of this! For example I didn't include any info from speed tests to avoid overwhelming amounts of info here. Thanks for your help.

Closes https://github.com/USEPA/EPATADA/issues/361, closes https://github.com/USEPA/EPATADA/issues/427, closes https://github.com/USEPA/EPATADA/issues/345, closes https://github.com/USEPA/EPATADA/issues/159


hillarymarler:
For TADA_FindNearby sites, the change that is causing the much longer run time is the includion of fetchNHD to only group sites that are within the same catchment. I have not tried this yet, but it would be possible to use the adjacency matrix approach only using the buffer distance, without grouping by catchment.

That is a potential option to prevent returning multiple groups for a single monitoring location and reduce the current run time. So I think the two options are really: 1) incorporate catchments (longer run time) or 2) rely only on buffer distance for grouping (shorter run time).

hillarymarler:
There are occasional test failures of ('test-ResultFlagsIndependent.R:137:3'): QC results are not flagged as Continuous ──
unique(cont_QC$TADA.ActivityType.Flag) == "Non_QC" is not TRUE
 
I am working on troubleshooting this one.

hillarymarler:
I am still seeing occasional failures of the continuous/QC results test. They seem to be related to TADA_RandomTestingData return empty data frames (which it should not do), so I'm taking a look at that now.

PR #586, opened 2025-04-08
Creating Module 3 Spatial Crosswalk Tables

wokenny13:
There are 3 new TADA Module 3 functions focused on helping users to apply unique site-specific criteria for assessments.

If a user is only interested in a ML sites level of assessments, they only need to run these 3 functions in order:

1) TADA_CreateParamRef()
2) TADA_CreateUseParamRef()
3) TADA_CreateSpatialRef()

Now, if a user is interested in an AU level of assessments, they can run two **additional** 'helper' functions prior to running TADA_CreateSpatialRef(). **See TADA Module 3b vignette for more info to walk through the process**:

4) TADA_CreateUseAURef()
5) TADA_CreateWaterUseParamRef()

This PR also reflects some minor updates to TADA_CreateParamRef() and TADA_CreateUseParamRef().

The excel file options in the functions still need updates, but the workflow should work when excel = FALSE


vignettes/TADAModule3_PartA.Rmd
```diff
@@ -344,13 +344,13 @@ Users may choose to voluntarily share their reference tables for their
 organization with the TADA Team so that others can use it in the future.
 
 ```{r}
-NCTC_ParamUseRef3 <- TADA_CreateUseParamRef(
+NCTC_UseParamRef3 <- TADA_CreateUseParamRef(
```
cristinamullin, line 484: This includes two rows for pH that are equivalent except for the TADA.ComparableDataIdentifier. This creates additional work since the use must be assigned to both even though they are both already assigned to the ATTAINS parameter "PH". Do we need TADA.ComparableDataIdentifier here if the ATTAINS parameter is already assigned as being the same for both?

wokenny13: I feel like there was a reason why TADA_ComparableDataIdentifier was left on here in TADA_CreateUseParamRef. But can't say for sure @hillarymarler do you recall any scenario in which TADA_ComparableDataIdentifier could be needed in TADA_CreateUseParamRef()?

I am unsure if there would ever be a case in which an ATTAINS parameter name, ex. if an organization is using ATTAINS Parameter name "Nitrogen", but it is only applied to a single use name for each TADA_ComparableDataIdentifier **"NITRATE + NITRITE_INORGANIC_AS NO3_MG/L", "NITRATE + NITRITE_UNFILTERED_AS N_MG/L", "TOTAL KJELDAHL NITROGEN (ORGANIC N & NH3)_UNFILTERED_AS N_MG/L"**
 whereas for all other use_names Nitrogen is assessed for only **"TOTAL KJELDAHL NITROGEN (ORGANIC N & NH3)_UNFILTERED_AS N_MG/L" and "NITRATE + NITRITE_INORGANIC_AS NO3_MG/L", "NITRATE + NITRITE_UNFILTERED_AS N_MG/L"**. 
image

However if TADA_ComparableDataIdentifier is needed in TADA_CreateUseParamRef, then would it be needed in TADA_DefineMagnitude?

R/ATTAINSCrosswalks.R
```diff
@@ -1239,24 +1292,24 @@ TADA_CreateUseParamRef <- function(.data, org_id = NULL, paramRef = NULL,
...
       "One or more organization identifiers entered by user is not found in ATTAINS. "
     ))
   }
 
   # Checks if org_id are found in the user supplied paramRef argument.
   if (sum(!org_id[org_id != "EPA304a"] %in% paramRef$organization_identifier) > 0) {
     warning(paste0(
-      "TADA_CreateParamRef: ",
-      "One or more organization identifiers entered by user is not found in your paramRef argument input.",
+      "TADA_CreateUseParamRef: ",
```
cristinamullin, line 1315: I wonder if this should be a "stop" instead of a "warning"? Because users cannot generate the ref for orgs missing from paramRef

wokenny13: I can change this to "stop"

R/CriteriaMethods.R
```diff
@@ -0,0 +1,436 @@
...
+#' function input. This helps prevent users from overwriting their progress.
+#'
+#' @param .data A TADA dataframe. Users should run the appropriate data cleaning,
+#' processing, harmonization and filtering functions prior to this step.
+#'
+#' @param spatialRef An optional data frame which contains the completed spatial
+#' crosswalk to assign any unique spatial criteria to a parameter, use, waterbody
+#' or monitoring site/assessment unit.
+#' 
+#' @param ref "TADA"
```
cristinamullin, line 40: Can you add a definition for the ref = "TADA" input and an example that uses it?

PR #595, opened 2025-05-21
used stringr::str_wrap solution

### LINKED ISSUE #594: TADA_InsertBreak() bug report
**Describe the bug**

`TADA_InsertBreaks()` gives "In min(which(spaces > x)) : no non-missing arguments to min; returning Inf" warning message for certain input strings and gives value error when string is NA. Consider replacing this with `stringr::str_wrap()` which does something similar and can handle instances when the string is NA.

**To Reproduce**

1\. Load EPATADA

2\. Run function

3\. See error

Code to reproduce the behavior:

``` r
library(EPATADA)

# runs ok
TADA_InsertBreaks(x = "Flag: MeasureQualifierCode indicates result value is suspect", len = 31)
# length values 31 and above do no generate error for this particular string

# Generates Inf warning
TADA_InsertBreaks(x = "Flag: MeasureQualifierCode indicates result value is suspect", len = 30)

# Generates missing value error
TADA_InsertBreaks(x = NA, len = 30)
```

**Expected behavior**

```
# load stringr
library(stringr)

# expected behavior should be like this
stringr::str_wrap(string = "Flag: MeasureQualifierCode indicates result value is suspect", width = 30)  # two returns added
stringr::str_wrap(string = "Flag: MeasureQualifierCode indicates result value is suspect", width = 5) # six returns added
stringr::str_wrap(string = NA, width = 30) # gives NA
```

**Screenshots**

Not applicable.

**Additional context**

I developed a partial solution, but this doesn't fully fix the issue because the return carriage is only being added after the first break instance; subsequent returns are not included. I'm having trouble getting to do what I want it to do but don't know if it's necessary to spend the time since there's an existing `stringr` function. Therefore, I think using the `stringr::str_wrap()` function would be best.

```
# partial modifications to the existing functions
TADA_InsertBreaks <- function(x, len = 50) {
  if (nchar(x) > len & !is.na(x)) { # added is.na() statement
    multiples <- floor(nchar(x) / len)
    lens <- seq(len, len * multiples, by = len)
    spaces <- unlist(gregexpr(" ", x))
    if (max(spaces) > len) {
      spots <- sapply(lens, function(x) spaces[min(which(spaces > len))]) # changed which() condition
      spots_unique <- unique(spots) # removed duplicate spots
      for (i in 1:length(spots_unique)) { # used length of unique spots
        stringi::stri_sub(x, spots[i] + (i - 1), spots[i]) <- "\n "
      }
    }
  }
  return(x)
}

# examples
insert_breaks(x = "Flag: MeasureQualifierCode indicates result value is suspect", len = 30) # only one return added
insert_breaks(x = "Flag: MeasureQualifierCode indicates result value is suspect", len = 5) # only first return added
insert_breaks(x = NA, len = 30) # gives NA
```

**Reminders for TADA contributors addressing this issue**

Bug fixes should include all the following work:

-   [x] Create or edit the function/code.

-   [x] Document all code using line/inline and/or multi-line/block comments
    to describe what is does.

-   [ ] Create or edit tests in tests/testthat folder to help prevent and/or 
    troubleshoot potential future issues.

-   [ ] Create or edit the function documentation. Include working
    examples.

-   [ ] Update or add the new functionality to the appropriate vignette
    (or create new one).

-   [ ] If function/code edits made as part of this issue impact other
    functions in the package or functionality in the shiny app, ensure
    those are updated as well.
    
-   [ ] Run TADA_UpdateAllRefs(), TADA_UpdateExampleData(), styler::style_pkg(),
    devtools::document(), and devtools::check() and address any new notes or 
    issues before creating a pull request.    

-   [ ] Run more robust check for releases: devtools::check(manual = TRUE, 
    remote = TRUE, incoming = TRUE)

hillarymarler:
Removed TADA_InsertBreaks and substituted stringr::str_wrap solution


hillarymarler:
Need to update some vignette branch refs to correct the check issues - doing that now.

cristinamullin (approved):

PR #618, opened 2025-07-07
Fix consolidated depth and depth flag bug

hillarymarler:
When a data set had no depth data available, the additional columns, TADA.DepthCategory.Flag and TADA.ConsolidatedDepth, were not added. This created problems in some workflows. This fix adds the two columns (filled with NA) when a data set has no depth data available.

### DIFF
--- a/R/DepthProfile.R
+++ b/R/DepthProfile.R
@@ -209,7 +209,13 @@ TADA_FlagDepthCategory <- function(.data, bycategory = "no", bottomvalue = 2, su
       dplyr::select(-ARD_Category, -DepthsPerGroup)
 
     if (depth.count == 0) {
-      print(paste("TADA_FlagDepthCategory: checking data set for depth values. No results have depth values available, TADA_FlagDepthCategory cannot be used on this data set.", sep = ""))
+      print(paste0("TADA_FlagDepthCategory: checking data set for depth values. No results have depth values available. ",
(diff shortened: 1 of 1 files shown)


wokenny13:
I made a minor update in my commit. This seems to resolve the problem for cases in which no depth data is available.

R/DepthProfile.R
```diff
@@ -207,13 +207,20 @@ TADA_FlagDepthCategory <- function(.data, bycategory = "no", bottomvalue = 2, su
         TADA.DepthCategory.Flag = ifelse(is.na(TADA.DepthCategory.Flag), "Not enough depth info to determine category", TADA.DepthCategory.Flag)
       ) %>%
       dplyr::select(-ARD_Category, -DepthsPerGroup)
+  }
+  
+  if (depth.count == 0) {
+    print(paste0("TADA_FlagDepthCategory: checking data set for depth values. No results have depth values available. ",
```
cristinamullin: Is paste needed here? 

Suggested update:
```r
print("TADA_FlagDepthCategory: No depth information was found in the dataset. The columns TADA.DepthCategory.Flag and TADA.ConsolidatedDepth are being added and populated with NA values.")
```

hillarymarler: I removed the unnecessary paste, @cristinamullin. Thanks!

PR #661, opened 2025-09-29
657 bot updates for maintenance tasks

### LINKED ISSUE #657: Bot updates for maintenance tasks
Workflow to run maintenance tasks and bot-update, e.g., 
```
# WQPWQXRefTables.R
TADA_UpdateWQXCharValRef()
TADA_UpdateMeasureUnitRef()
TADA_UpdateDetCondRef()
TADA_UpdateDetLimitRef()
TADA_UpdateActivityTypeRef()
TADA_UpdateCharacteristicRef()
TADA_UpdateMeasureQualifierCodeRef()
TADA_UpdateMonLocTypeRef()
TADA_UpdateWQPOrgProviderRef()
```

**Unknowns**
Are the source materials for each of these updated: 

- (A) on a set schedule (e.g., file updated nightly), allowing these to be a scheduled task, or
- (B) intermittently (e.g., as data are added to WQX), meaning these would need to be triggered by a failed check (i.e., pre-commit)

Assuming A: How much oversight do we want on the commits (e.g., maintainer review each update as a PR or just let the bot PR to develop?)

jbousquin:
Added maintain.yml and all other file changes are done by bot commits. See issue #657 for links to specific workflow runs and options.

Note: By design this may have merge conflicts because the branch was before a merge that updated reference files to ensure these workflows would bot-update those files.


.github/workflows/maintain.yaml
```diff
@@ -0,0 +1,246 @@
...
+        with:
+          use-public-rspm: true
+
+      - name: Install dependencies
+        uses: r-lib/actions/setup-r-dependencies@v2
+
+      - name: Update reference files
+        run: |
+          install.packages('remotes')
+          remotes::install_github("USEPA/TADAShiny", ref = "develop", dependencies = TRUE, force = TRUE)
```
cristinamullin: This should install USEPA/EPATADA here not USEPA/TADAShiny

jbousquin:
actions/checkout@v4 should be checking out the 'develop' branch, not sure why it is finding itself in a detached head state. Possible because merging is blocked? I'll try being more direct with what to checkout.

jbousquin:
Running this on a PR works, hoping once this is merged scheduled runs do most of the work. FYI on timing and current sticking point:
- On initiation of the PR and subsequent commits it'll run the updates at the same time as tests.
- If no changes it'll run checks just like normal.
- If it finds changes to make it'll bot-commit those to the PR.
The current sticking point is for it to re-run checks after the bot-commit. github recognizes the default github_token the workflow is set up with and ignores the new commit for CI runs. It does that to avoid circular CI (where a check would commit, cause itself to run again, etc.). Ideally we want it to run checks only after the workflow completes and not re-run itself. If we have to use a token I'd rather it be at the repo level not associated with a specific admin user.

.github/workflows/maintain.yaml
```diff
@@ -0,0 +1,246 @@
...
+        with:
+          use-public-rspm: true
+
+      - name: Install dependencies
+        uses: r-lib/actions/setup-r-dependencies@v2
+
+      - name: Update reference files
+        run: |
+          install.packages('remotes')
+          remotes::install_github("USEPA/EPATADA", auth_token = Sys.getenv("auth_token"), ref = "develop", dependencies = TRUE, force = TRUE)
```
cristinamullin: you may need to add library(remotes) after the remotes install here

jbousquin:
All my edits since my last comment were to try to work around this... copying here for easy reference:

Because this is using the built-in [GITHUB_TOKEN secret](https://docs.github.com/en/actions/security-guides/automatic-token-authentication): GitHub sees that the push event has been triggered by a commit generated by CI, and doesn't run any further checks to avoid unintentional check loops.

To have commits generated during CI trigger other workflow runs, you can checkout the repo using a [Personal Access Token (PAT)](https://docs.github.com/en/authentication/keeping-your-account-and-data-secure/creating-a-personal-access-token): this will make the resulting commit the same as if you made it yourself. If using actions/checkout, like we are, their [docs](https://docs.github.com/en/authentication/keeping-your-account-and-data-secure/creating-a-personal-access-token) suggest how to set a repo token.

jbousquin:
### Action items on this
Split workflow triggers:
1. to generate PR when run on schedule/manual run (as new [branch](https://github.com/USEPA/EPATADA/tree/657-bot-updates-scheduled)/[PR](https://github.com/USEPA/EPATADA/pull/663))
2. to run on submitted PR (will stay here)

For submitting on PR - try either as a pre-commit or to use an action to pause other checks until after the new commit (will have to make sure the check is run on correct point in branch)

jbousquin:
Workflow failure confirmed anticipated issues:

1. The updated files are not added to the cache, and
2. No .pre-commit-config.yaml

Will pursue use of other actions next. Note: pre-commit workflow should be easier to figure out with formatter/linting, especially if written in rust ([air](https://www.tidyverse.org/blog/2025/02/air/)).

cristinamullin (approved):

.github/workflows/maintain.yaml
```diff
@@ -0,0 +1,230 @@
...
+        run: |
+          install.packages("remotes")
+          library(remotes)
+          remotes::install_github("USEPA/EPATADA", ref = "develop", dependencies = TRUE, force = TRUE)
+          library(EPATADA)
+          # ATTAINSRefTables.R
+          EPATADA:::TADA_UpdateATTAINSOrgIDsRef()
+          EPATADA:::TADA_UpdateATTAINSParamUseOrgRef() # takes a long time
+          EPATADA:::TADA_UpdateATTAINSParameterWQPCharRef()
+          # WQPWQXRefTables.R
```
cristinamullin: Do you think we should I create a single maintenance function for use here that includes all of these functions? In addition, I will add the separate function for updating all example data that we discussed.

jbousquin: We could - that would help reduce the size of this file for sure and make it easier to add/remove functions over time.

jbousquin:
On a PR in another repo ([harmonize-wq](https://github.com/USEPA/harmonize-wq/pull/128)) I was able to confirm even if the token is not specified as part of checkout it still just commits changes without re-running checks after (specific [commit](https://github.com/USEPA/harmonize-wq/pull/128/commits/98ac7527ef5562b685919125c758cd3516121961)).

What to try next:

1. Figure out as pre-commit for format/lint (new issue #674), and then see if that approach is a viable solution here
2. Add [action](https://github.com/lewagon/wait-on-check-action) to hold checks until after commit
3. Set up a bot token to get it to still run the checks

cristinamullin:
@jbousquin what are your thoughts on next steps for this PR? I wonder if might be best to set it up similar to [format-batch-update.yaml](https://github.com/USEPA/EPATADA/blob/develop/.github/workflows/format-batch-update.yaml) for now? Then it could be run on a branch as needed. I think that alone might save us a lot of time.

jbousquin:
You can manually trigger [maintenance-scheduled.yaml](https://github.com/USEPA/EPATADA/blob/develop/.github/workflows/maintenance-scheduled.yaml) on your branch/PR to accomplish similar but I think that would commit the change to the maintenance-branch instead. Will play around with/put some thought into that. Also OK with closing this PR if we're happy with just the maintenance bot.

jbousquin:
Updating this to run manually on branch before PR (like format-update).

jbousquin:
Updating scheduled to no longer run manually (help keep it clear what to run when)

PR #663, opened 2025-10-02
657 bot updates scheduled

jbousquin:
Initial attempt to add scheduled workflow for file maintenance tasks.

### DIFF
--- a/.github/workflows/maintenance-scheduled.yaml
+++ b/.github/workflows/maintenance-scheduled.yaml
@@ -1,0 +1,247 @@
+name: Scheduled Component File Update
+
+on:
+  # Daily at 2:01AM https://crontab.guru/
+  schedule:
+    - cron: '1 2 * * *'
+  # Allows you to run this workflow manually from the Actions tab
+  workflow_dispatch:
+
+jobs:
+  maintenance_updates:
+    runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


jbousquin:
This will need to be merged before we can really do much more testing.

Update: Was able to confirm on a fork that this workflow (triggered from workflow_dispatch on develop) will create a branch from where we want (origin/develop), create a second commit on the same branch if needed for example data, but will not create a PR for that branch when it is done (just the branch).

jbousquin:
Was able to further confirm later commits will be made to the same branch, i.e., if we don't get around to merging a branch from a scheduled run before it runs again the commit will be made on the existing branch without issue.

cristinamullin:
Is it possible to make this create a PR from the branch with the maintenance updates automatically? Will this run weekly or still daily?

jbousquin:
> Is it possible to make this create a PR from the branch with the maintenance updates automatically?

Yes, scoping how much of a lift that will be.

> Will this run weekly or still daily?

Currently set to run nightly, figured that would be helpful to troubleshooting in the next 24hours. Also note that it won't create a new branch, just continues to update. Easy to change it to weekly but would need to know what day of the week?

cristinamullin (approved):

PR #683, opened 2025-10-16
Maintenance branch

jbousquin:
Periodic updates.
Note: I was looking at some at some of the files updated and it looked like some of the changes might just be changed order - how much are these results sorted before saving?


cristinamullin:
> Periodic updates. Note: I was looking at some at some of the files updated and it looked like some of the changes might just be changed order - how much are these results sorted before saving?

We order the columns - this runs within most functions: [TADA_OrderCols()](https://usepa.github.io/EPATADA/reference/TADA_OrderCols.html)

So that ordering would happen for the example data produced. I don't think we order the rows, though they could potentially be ordered by result ID if you think that would be helpful?

For the reference files, I don't believe we do any ordering.

jbousquin:
I'd trust the reference data coming in (response from the service) to have more consistent column than row order. The data will be the same, just hopefully fewer updates that aren't real changes. More specifically - seeing this in the csv files. Will generate issue for it.

Update: #684

cristinamullin (approved):

cristinamullin (approved):

PR #687, opened 2025-10-16
680 duplicate function tada aggregatemeasurements

### LINKED ISSUE #680: Duplicate function? (TADA_AggregateMeasurements)
The function TADA_AggregateMeasurements if defined in both Transformations.R and Utilities.R. They seems to be very similar. Can it be safely removed from Transformations.R?

jbousquin:
Kept function in Transformations, dropped function in Utilities, updated docs.


.github/workflows/R-CMD-check.yaml
```diff
@@ -2,7 +2,8 @@
 # Need help debugging build failures? Start at https://github.com/r-lib/actions#where-to-find-help
 on:
   push:
-    branches: develop
+    branches:
+      [develop]
   pull_request:
     branches: develop
```
cristinamullin, line 8: does this change need to be made on this line as well?

[develop]

jbousquin: If anything I'd say to revert it to a one-liner for push_request - I left it as an extra line/list to make it easier for myself to run CI on push to future branches, don't usually do PRs to other branches so I left that as the better one-liner.

cristinamullin (approved):

Issue #690, opened 2025-10-17
TADA_ThresholdFlag


jbousquin:
TADA_FlagBelowThreshold() [615-816] & TADA_FlagAboveThreshold() [408-611] are very similar, just with one comparing on < and the other >. Each function should do one thing well - but this seems like an opportunity to reduce duplicative code. The other thing that isn't clear to me is how to flag both - as when I flag one it drops the existing flag column.

'Flag' is also a bit of a misnomer being that the function will flag/filter/clean. 

One option would be an arg with 3 acceptable params [Below/Above/Between]. Maybe shorten to TADA_Thresholds()? Not sure how R does with that since the boolean logic makes me think you have tried to avoid args like that?

Another option would be to add an internal function that does all the work and keep the current two as is, passing the internal function info specific to above/below, thus letting users interact with it the way it is (v1.0 suggests some stability).

PR #699, opened 2025-10-17
Range flag update

cristinamullin:
The upper and lower flag functions were not accounting for invalid units. If the units are not valid, then the ranges are not either and now those rows are also flagged as suspect. 

TADA_FlagAboveThreshold
TADA_FlagBelowThreshold


tests/testthat/test-Utilities.R
```diff
@@ -187,49 +187,57 @@ test_that("pH harmonization works as expected throughout workflow", {
...
-
+  
   # Process data
   # - Applies several functions to clean and harmonize the data.
   ph_data <- ph_data %>%
     TADA_SimpleCensoredMethods() %>%
     TADA_ConvertSpecialChars(col = "TADA.ResultMeasureValue", clean = TRUE) %>%
     TADA_RunKeyFlagFunctions(clean = TRUE) %>%
     TADA_HarmonizeSynonyms()
-
```
jbousquin, line 228: This looks like it's adding whitespace back in? (i.e., a blank line being removed to add a line with '  ')

tests/testthat/test-Utilities.R
jbousquin, line 1: I like the try/catch for TADA_DataRetrieval within this test, is there a way to do it more widespread?

jbousquin (approved):
Looks like an old (>1 year) branch that was brought forward with some lingering changes (e.g., Data_Nutrients_UT -> Data_R5_TADAPackageDemo) that I don't have the context to ensure are intentional vs a revert that made it through merges.

I do see where the bad units handling was added to the two functions in ResulFlagsIndependent and it seems like it should accomplish what it is intended to.

The couple of in-line comments I made were whitespace formatting, a quick styler() run should fix throughout (approving without requiring changes to expedite).

jbousquin:
Note that this makes these two functions longer while doing the same thing across both (relevant to #690)

tests/testthat/test-Utilities.R
cristinamullin: Good idea. I am working on adding better error messaging and a max_attempts function input to TADA_DataRetrieval with the default of 3.

cristinamullin:
> Looks like an old (>1 year) branch that was brought forward with some lingering changes (e.g., Data_Nutrients_UT -> Data_R5_TADAPackageDemo) that I don't have the context to ensure are intentional vs a revert that made it through merges.
> 
> I do see where the bad units handling was added to the two functions in ResulFlagsIndependent and it seems like it should accomplish what it is intended to.
> 
> The couple of in-line comments I made were whitespace formatting, a quick styler() run should fix throughout (approving without requiring changes to expedite).

I ran styler to remove extra white space/blank lines. The switch from Data_Nutrients_UT -> Data_R5_TADAPackageDemo is intentional in the documentation for the flag functions. Data_Nutrients_UT is already somewhat clean (some functions have already been run on it) so it is not a good example for those.

PR #708, opened 2025-11-19
Vignette Re-ordering

mhweber:
- Reorder vignettes by restructing _pkgdown.yml

### DIFF
--- a/_pkgdown.yml
+++ b/_pkgdown.yml
@@ -1,4 +1,27 @@
-url: "usepa.github.io/EPATADA/"
+url: usepa.github.io/EPATADA/
 template:
   bootstrap: 5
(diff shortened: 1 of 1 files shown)


cristinamullin (approved):

_pkgdown.yml
```diff
@@ -1,4 +1,27 @@
-url: "usepa.github.io/EPATADA/"
+url: usepa.github.io/EPATADA/
 template:
   bootstrap: 5
   bootswatch: yeti
+  light-switch: true
+articles:
+- title: Articles
```
cristinamullin, line 7: @hillarymarler @wokenny13 
FYI - if the names change or if we add new ones, we should now change them here too.

PR #711, opened 2025-11-21
DO NOT MERGE - Mock edit

### LINKED ISSUE #674: Pre-commit formatting/linting
> New features and/or edits should include all the following work:
> -   [ ] Run TADA_UpdateAllRefs(), TADA_UpdateExampleData(), **<ins>styler::style_pkg()</ins>**,
>     devtools::document(), and devtools::check() and address any new notes or 
>     issues before creating a pull request.

**Describe the solution you'd like:**

We should be able to make that part of PR pre-commits. This would be either as a check (my preference) or where it makes changes directly. 

**Additional context:**

Stay w/ styler package or transition to [air](https://www.tidyverse.org/blog/2025/02/air/#how-is-this-different-from-styler)?

Stay with default code format/style rules, and if not what to configure/enforce?
For example, the [tidyverse style guide](https://style.tidyverse.org/pipes.html) uses the native pipe by default for new code (something we're trying to do moving forward - issue #642 ). It's not clear to me what using the tidyverse_style() arg vs some of the other CI (e.g., [usethis::use_tidy_style()](https://usethis.r-lib.org/reference/tidyverse.html)) enforces.

I'm familiar with [pre-commit/action](https://github.com/pre-commit/action), looks like [styler suggests](https://styler.r-lib.org/articles/third-party-integrations.html) [lorenzwalthert/precommit](https://github.com/lorenzwalthert/precommit)

jbousquin:
Shortened comment - this isn't meant to be merged, just testing formatting checks

### DIFF
--- a/R/MaintenanceScheduled.R
+++ b/R/MaintenanceScheduled.R
@@ -42,7 +42,7 @@
   tryCatch(
     {
       TADA_UpdateATTAINSOrgIDsRef()
-      TADA_UpdateATTAINSParamUseOrgRef() # takes a long time
+      TADA_UpdateATTAINSParamUseOrgRef() # slow
       TADA_UpdateATTAINSParamToWQPCharRef()
     },
     error = function(e) {
--- a/R/MaintenanceScheduled.R
+++ b/R/MaintenanceScheduled.R
@@ -160,7 +160,7 @@
         TADA_FindPotentialDuplicatesSingleOrg() %>%
         dplyr::filter(!(MeasureQualifierCode %in% c("D", "H", "ICA", "*"))) %>%
         TADA_SimpleCensoredMethods(
-          nd_method = "multiplier",
+           nd_method = "multiplier",
           nd_multiplier = 0.5,
           od_method = "as-is",
           od_multiplier = "null"


jbousquin:
Note in the example [commit](https://github.com/USEPA/EPATADA/pull/711/commits/82bba09c74290e14937aafa0d9ce1403b4b36c01) where the changed line of code had no format problems no suggestions were made and the [check passed](https://github.com/USEPA/EPATADA/actions/runs/19574489411/job/56056092065?pr=711)

jbousquin:
Note the second [commit](https://github.com/USEPA/EPATADA/pull/711/commits/6be58392c712102fb717e08e57dde7a241646704) purposely changed indentation in a line of code giving it the wrong leading whitespace. The check failed and a comment suggestion was posted to the PR to show how to fix the whitespace.

jbousquin:
The defaults (see [configuration](https://posit-dev.github.io/air/configuration.html)) seem good for now, and it really doesn't enforce much:

- line-width = 80
- indent-width = 2
- indent-style = "space"
- line-ending = "auto"
- persistent-line-breaks = true

The only thing I might change is persistent-line-breaks. IMHO you _should_ flatten to one line when it will fit within the line-width (fewer lines of code), but there are exceptions to every rule and this configuration will allow either, empowering the author to write it the way they think is best.

To skip formatting on a line of code use:
`# fmt: skip`
I would anticipate that'll happen most often to ignore line-width, as there are times when you want a string (e.g., url) to be one line.

cristinamullin:
> The defaults (see [configuration](https://posit-dev.github.io/air/configuration.html)) seem good for now, and it really doesn't enforce much:
> 
> * line-width = 80
> * indent-width = 2
> * indent-style = "space"
> * line-ending = "auto"
> * persistent-line-breaks = true
> 
> The only thing I might change is persistent-line-breaks. IMHO you _should_ flatten to one line when it will fit within the line-width (fewer lines of code), but there are exceptions to every rule and this configuration will allow either, empowering the author to write it the way they think is best.
> 
> To skip formatting on a line of code use: `# fmt: skip` I would anticipate that'll happen most often to ignore line-width, as there are times when you want a string (e.g., url) to be one line.

I agree we should set persistent-line-breaks = **false**. On the most recent PR's we are running into an issue with this workflow where it is making many separate comments that can be manually added to a "batch" commit. It will be much simpler/better for our dev team if they can be automatically added to a batch commit made directly to the PR/branch. Then the team can review the changes in that commit before merging the work into develop. At that time, if needed (very rare), the commit could be reverted.

PR #721, opened 2025-12-03
Batch air format workflow (and applied to existing code base)

jbousquin:


jbousquin:
Mentioned on call: failing test-coverage check is because this PR is from a fork, it runs fine but posting the comment back on the PR is intentionally disabled since that would required write-permissions that could allow the workflow to be altered to create a vulnerability. The coverage report is available as an artifact on the action instead: (https://github.com/USEPA/EPATADA/actions/runs/19911976638/artifacts/4765708233)
Not as easy to review, but rare we run into it.

To be clear this is unrelated to this PR (see instead #722), just background why you can ignore the failed check.

For reference see [cobertura-action](https://github.com/5monkeys/cobertura-action) and [github-actions-preventing-pwn-requests](https://securitylab.github.com/resources/github-actions-preventing-pwn-requests/).

cristinamullin (approved):

PR #727, opened 2025-12-08
Update TADA_FlagResultUnit

cristinamullin:
TADA_FlagResultUnit should only flag results by default. In addition, users should only be able to return only flagged results if clean = "none". This bug has been fixed. Other minor updates were made to improve documentation.


R/GeospatialFunctions.R
```diff
@@ -2381,9 +2381,14 @@ TADA_GetATTAINSByAUID <- function(
     df <- .data %>%
       dplyr::left_join(geo.data, by = c("ATTAINS.AssessmentUnitIdentifier"))
 
-    # bind with existing attains.geo data
-    attains.geo <- plyr::rbind.fill(attains.geo, df) %>%
-      dplyr::filter(!is.na(GLOBALID))
+    # Bind with existing attains.geo data
```
cristinamullin, line 2384: @hillarymarler I noticed there was an issue when the df did not contain GLOBALID. This is a quick attempt to work around that issue. Flagging as an FYI - Not sure if we might need to address this elsewhere.

R/MaintenanceScheduled.R
```diff
@@ -380,6 +378,7 @@
       # =======================================
       # Generate Data_MT_UseAURef
       # =======================================
+      Data_MT_AUMLRef <- Data_MT_AUMLRef$ATTAINS_crosswalk
```
cristinamullin, line 381: @wokenny13 I was running into an error when trying to re-create these df's. You may have already addressed this in your PR... flagging as an FYI

wokenny13: I think the combined-review-cm branch should now address the errors. All the example data sets should be able to run fine now on that branch

PR #751, opened 2025-12-18
Maintenance branch

jbousquin:
Re-attempt at #750 , being lazy and didn't want to deal with merge-conflicts. May be a reason to keep the scheduled workflow able to be run manually.

Downside is by deleting the existing you lose how many commits were may by the scheduled task (~39), which is only useful for 'has it been a long time since we merged those?'.


inst/extdata/WQXMeasureQualifierCodeRef.csv
```diff
@@ -125,7 +125,7 @@
 "Result Measure Qualifier(MeasureQualifierCode)",827,"HQ","High Quality control measures","4/10/2020 6:07:22 PM","Pass"
 "Result Measure Qualifier(MeasureQualifierCode)",158,"HTH","Hard to Homogenize","3/16/2016 1:04:25 PM","Pass"
 "Result Measure Qualifier(MeasureQualifierCode)",1194,"HTI","Holding time for test is immediate; the laboratory measurement, therefore, cannot be used for compliance purposes","7/18/2024 8:49:50 PM","Not Reviewed"
-"Result Measure Qualifier(MeasureQualifierCode)",229,"HV","Result value is defined as highly variable by data owner. (HV) High variability: questionable precision and accuracy","12/11/2025 6:37:41 PM","Not Reviewed"
+"Result Measure Qualifier(MeasureQualifierCode)",229,"HV","High variability: questionable precision and accuracy","12/18/2025 8:33:22 AM","Not Reviewed"
```
cristinamullin, line 128: This is an example of an additional TADA maintenance need that is not yet documented... for now I will add a note about it to the requiredMaintenance.R

In this case, a new value was added to the WQX measure qualifier domain and the TADA team has not yet categorized it as "Suspect", "Pass", etc. By default, the code that produces WQXMeasureQualifierCodeRef.csv assigns any new domains the category, "Not Reviewed". In this case, the relevant function is TADA_GetMeasureQualifierCodeRef() and the code starts on line 720 in https://github.com/USEPA/EPATADA/blob/develop/R/WQPWQXRefTables.R. What I need to do, or what anyone maintaining the package in the future would need to do, is update the TADA_GetMeasureQualifierCodeRef() to add "HV" to the relevant list. In this case, I will add this one to the "suspect" list in the code.

jbousquin: Am I understanding correctly that this is about maintaining WQPWQXRefTables.R? Or is it maintaining how it gets/uses that info in TADA_GetMeasureQualifierCodeRef()?

Edit (I was not understanding): Looks like our csv [inst/extdata/WQXMeasureQualifierCodeRef.csv](https://github.com/USEPA/EPATADA/pull/751/files/fcd340ba0217d2cb600b39b2882ff0dfda130eda#diff-fbd80f0c9ccff36f4d8984a75baf6dcc3813acea1fbccb5f341ce21e111cf91f) that this maintained by this workflow is derived from wqx [ResultMeasureQualifier.CSV](https://cdx.epa.gov/wqx/download/DomainValues/ResultMeasureQualifier.CSV), and our function that you referenced has additional information, specifically a list of e.g., suspect to add the additional column? So the additional maintenance is on the function to produce that file where we need to review them/assign to correct list? I could see that as part of periodic maintenance (e.g., on release) but not part of pre-PR.

cristinamullin: This is about needing to maintain the TADA_GetMeasureQualifierCodeRef() function in WQPWQXRefTables.R which produces the WQXMeasureQualifierCodeRef.csv for use in the TADA package. That ref is used for easy filtering using the TADA_FlagMeasureQualifierCode() function. 

TADA_GetMeasureQualifierCodeRef() first creates a copy of the WQX domain table (https://cdx.epa.gov/wqx/download/DomainValues/ResultMeasureQualifier.CSV). It is not just a copy though... the function then adds a column to categorize the codes for easy filtering. The problem is... new codes get added to WQX over time... so then the TADA team has to go back to edit the TADA_GetMeasureQualifierCodeRef() function and add the category.

inst/extdata/AKAllotments.dbf
cristinamullin, line 1: All of the tribal .dbf files (AKAllotments, AKVillages, AmericanIndian, OKTribe, OffReservation, VATribe) probably don't need to be updated often. Right now, the files are replaced each time TADA_UpdateTribalLayers() is run regardless of if something changed... there is actually no way to know if anything actually did change in the file itself. I am not sure if there is a last change date that could be leveraged so they only are only updated if changes were actually made. If there is, we could update: TADA_UpdateTribalLayers() to achieve this.

inst/extdata/WQXMeasureQualifierCodeRef.csv
jbousquin: > I will add this one to the "suspect" list in the code.

Are you looking to do that as part of this PR or a new one? I'm hoping to just merge this to get #661 passing

inst/extdata/AKAllotments.dbf
jbousquin: Added #753 to address this

inst/extdata/WQXMeasureQualifierCodeRef.csv
cristinamullin: Added https://github.com/USEPA/EPATADA/issues/754 to create a prompt for this in the future

cristinamullin (approved):

PR #760, opened 2025-12-19
Add GitHub Actions workflow for spell checking

### LINKED ISSUE #759: Make spelling::spell_check_package() part of CI actions
I have ideas of how to do this, leaning toward adding it to Component File Update over format-update. Could also be its own but a lot of set-up involved for such a small thing.

jbousquin:
This was initially started in response to this PR template check:
-   [ ] Run spelling maintenance in `requiredMaintenance.R`.

We can run the check as part of CI, but it would be a bad idea to automatically add any words to the WORDLIST in case the contributor didn't do what they should have. Instead this workflow is a check, i.e., to let reviewers know if the contributor didn't do what they should have.

Note: the spelling library can be used as a check within R-CMD-check, but:
(1) that isn't enabled currently or by default,
(2) I feel like it would get lost in there, and 
(3) that workflow already takes a long time


jbousquin:
This PR has the workflow run on it allowing you to see what it looks like when passing (1min 34s):
[Passing spelling-check](https://github.com/USEPA/EPATADA/actions/runs/20383773717/job/58580270698?pr=760)
However, I wanted to demo what a failure looks like. Temporarily I set that up on a fork (1min 26s):
[Failed spelling-check](https://github.com/jbousquin/EPATADA/actions/runs/20383531657)

cristinamullin (approved):

PR #775, opened 2026-01-08
Refactoring and maintenance

cristinamullin:
### Pull Request Checklist

Preparation

-   [ ] Update your branch from the latest `develop` and resolve any merge conflicts

-   [ ] Before creating a pull request trigger the [format-update](https://github.com/USEPA/EPATADA/actions/workflows/format-batch-update.yaml) GitHub Action on your branch to format the code

Documentation

-   [ ] Add/update inline and/or block comments to clarify complexity, context and intent

-   [ ] Add/update function documentation (roxygen), include working examples, build docs locally using devtools::document(), and inspect added/updated help pages for content and format

-   [ ] Add/update vignettes for corresponding changes in functionality, list these under articles in _pkgdown.yml, and ensure added/updated vignettes run and build with proper formatting locally

Maintenance & Data Refresh 

-   [ ] Add new dependencies to `DESCRIPTION` and document appropriately

-   [ ] Run spelling maintenance in `requiredMaintenance.R`

-   [ ] If changes affect other package or the shiny app functions, update those impacted functions accordingly

-   [ ] If columns were added/updated, add/update them in `RequiredCols.R`

-   [ ] Run `.TADA_UpdateRefFiles()` and `.TADA_UpdateExampleData()` locally via `MaintenanceScheduled.R` or trigger the [Component File Update](https://github.com/USEPA/EPATADA/actions/workflows/maintenance-update.yaml) GitHub Action

-   [ ] If new example data files were added, document them in `ExampleData.R` and include them in `MaintenanceScheduled.R` for regular refresh

Tests and checks

-   [ ] Add/update tests in `tests/testthat` to cover changes

-   [ ] Run `devtools::test()` to verify new and existing tests pass

-   [ ] Run `devtools::check()` and address any errors, warnings or notes

Pull Request Description

-   [ ] Includes a summary of the changes made

-   [ ] Includes relevant context/motivation

-   [ ] Includes links to related issues or pull requests (keywords like "Closes #issue_number" automatically close related issues when pull request is merged)

Review

-   [ ] Review the bot-commented coverage-report generated by [test-coverage](https://github.com/USEPA/EPATADA/blob/develop/.github/workflows/test-coverage.yaml) to confirm all changes are covered by tests

-   [ ] Review/accept any bot-suggested format changes
  
-   [ ] Request review from at least one developer team member

PR #780, opened 2026-01-14
692 coverage geospatialfunctions

### LINKED ISSUE #692: Coverage - GeospatialFunctions.R
This seemed like a good one to start with given planned work on #636

jbousquin:
### Add test for fetchATTAINS() where coverage missing
Starting by just adding a test to run the function and not worrying about asserting actual result == expected result. 

Test data were added to use as inputs. This data does not need to be updated and probably shouldn't be (unless there is an edge case added in the future). It may be able to be further reduced but is relatively small.

### Pull Request Checklist

Preparation

-   [x] Before creating a pull request trigger the [format-update](https://github.com/USEPA/EPATADA/actions/workflows/format-batch-update.yaml) GitHub Action on your branch to format the code

Documentation

-   [x] Add/update inline and/or block comments to clarify complexity, context and intent

NA - Add/update function documentation (roxygen), include working examples, build docs locally using devtools::document(), and inspect added/updated help pages for content and format

NA - Add/update vignettes for corresponding changes in functionality, list these under articles in _pkgdown.yml, and ensure added/updated vignettes run and build with proper formatting locally

Maintenance & Data Refresh (NA)

Tests and checks

-   [x] Add/update tests in `tests/testthat` to cover changes

-   [x] Run `devtools::test()` to verify new and existing tests pass

-   [x] Run `devtools::check()` and address any errors, warnings or notes

Pull Request Description

-   [x] Includes a summary of the changes made

-   [x] Includes relevant context/motivation

-   [x] Includes links to related issues or pull requests (keywords like "Closes #issue_number" automatically close related issues when pull request is merged)

Review

-   [x] Review the bot-commented coverage-report generated by [test-coverage](https://github.com/USEPA/EPATADA/blob/develop/.github/workflows/test-coverage.yaml) to confirm all changes are covered by tests

-   [x] Review/accept any bot-suggested format changes
  
-   [x] Request review from at least one developer team member


jbousquin:
This bumped coverage from ~27% up to 34%, specifically targeting lines 341-566. This is passing, but I want to be clear the result we're currently getting isn't what we should.

Update (expected number of rows from layers based on python code):
points (lyr0) = 0
lines (lyr1) = 11
polygons (lyr2) = 1
catchments (lyr3) = 1318

I can upload these as expected, but this will be failing until resolved (really as it should).

jbousquin:
@hillarymarler This is the same error you mentioned on PR #775, just tried to break it out. Did you get anywhere toward resolving before I start digging in more on it?

jbousquin:
When I went step-wise, treating this large request like one that was < the area threshold, I had a local dependency versioning issue the try/catch in fetch_bbox() was suppressing. Once resolved I'm getting the expected 1318 catchment features from fetch_bbox().

When running fetchATTAINS() with the large request conditional commented out I'm getting:
points (lyr0) = 0
lines (lyr1) = 10
polygons (lyr2) = 1
catchments (lyr3) = 59
This checks out since they are likely clipped to the intersecting points from all returned within the bbox.

Recent commits comment out the large area handling to force this to run the tested way. Once that matches expected (passing) we can have more confidence in that expected being the target result from large area handling.

NOTE: formatting suggestions are on commenting around whitespace which I'm trying to keep as-is to make it easier to revert.

jbousquin:
Not sure why it is getting 59 catchments locally but 46 in CI - but CI is consistent and local dependency versioning is less than ideal.

jbousquin:
Tests are passing when large area is skipped, now to uncomment large area and see if it fails (should get same results)

jbousquin:
Ironically it runs fine but is getting 59 catchments

jbousquin:
Passing, only difference between normal bbox workflow and large area is more catchments come back for the latter. We'll want to pull those results in and compare against the points to confirm which is the better result.

jbousquin:
Map
Exported the input points (green), and result catchments (blue), as geopackages and mapped them alongside catchments directly from the service (orange outline).

The inputs and results are in the same CRS and a quick spatial selection confirmed all catchments results intersect an input point. This suggests the discrepancy (46 vs 59) is caused by a problem in the non-large area handling section of code.

jbousquin:
### Summary for review
This is a small PR adding coverage for the fetchATTAINS() large area handling. It includes:

1. input data for the test  (.rda),
2. a line of code to read that data in as a fixture for tests (commented out how it was retrieved), and
3. the actual test, with assertions that the expected number of features came back from each of the four services

This successfully increases coverage by ~7% for the R file containing it.

#### Next steps/New Issues:
Outside of this new test we only test that the function doesn't error when the catchments_only argument is TRUE/FALSE, and that if both those return results legnth(FALSE results)>length(TRUE results). This should always pass as long as the correct structure is returned even if there are 0 points/lines/polygons. I noticed a different number of results were coming back when not using the large area handling, #782 was created to determine what is happening differently in that part of the code and add tests to make sure future refactoring doesn't negatively impact results.

#### Suppressing warnings/errors:
Initially I was getting no results back from these services when run locally. I was able to determine there was an outdated dependency (rlang) version that was causing an issue and needed to be updated, but my concern is the error was being suppressed by:

```R
features <- suppressMessages(suppressWarnings({
        tryCatch(geojsonsf::geojson_sf(url(query)), error = function(e) NULL)
      }))
```

The way we currently page through results I think this is meant to catch the error where each page of results in incomplete. When we refactor we should address it (started to on an old branch so I have suggestions).

#### Maintenance Note:
The input data should not be updated. The test will however fail if the ATTAINS services change in a way that would alter the number of features that come back from the function. This seemed like the least sensitive test (e.g., doesn't test that fields/values stay consistent).

tests/testthat/test-GeospatialFunctions.R
```diff
@@ -64,6 +74,17 @@ testthat::test_that("fetchATTAINS fails with appropriate errors", {
   )
 })
 
+testthat::test_that("fetchATTAINS handles large areas", {
+  # large_bbox_data from fixtures (testdata/Hill_MT_pH.Rd)
+  testthat::expect_no_error(
+    result_all_features <- EPATADA:::fetchATTAINS(.data = large_bbox_data)
+  )
+  expect_equal(nrow(result_all_features$ATTAINS_points), 0)
```
hillarymarler, line 82: Will this need to be changed if any updates to the Hill_MT_ph.Rd example data set yield different results?  (Or is the plan that this example data will not be updated as part of regular maintenance?) 

For example, if a new monitoring location associated with an assessment unit not previously part of this example, the number of expected rows would change.

hillarymarler (approved):

tests/testthat/test-GeospatialFunctions.R
jbousquin: The input data (Hill_MT_ph.Rd) should not be updated.

Issue #841, opened 2026-02-11
Reduce test run-time for GeospatialFunctions


jbousquin:
This bridges both the parent issue and #808. 

Running tests are 20-30min, which is slow considering we're still adding tests. We're taking the proper approach as we add tests, starting with whatever data is handy. While that can mean extra/repeated processing, we're balancing it against our time and effort. It'll be useful to identify those tests that take the longest and then target them for more efficient test cases/data.

Of the slowest 10 tests, 3 are currently for Geospatial Functions (from `devtools::test(reporter = "slow")`:
[282.75s] test-GeospatialFunctions.R: TADA_FindNearbySites returns expected metadata
[264.31s] test-GeospatialFunctions.R: TADA_FindNearbySites returns expected number of site groups
[120.81s] test-GeospatialFunctions.R: fetchATTAINS handles large areas

An example of overlap w/ refactoring is fetchATTAINS handles large areas, where we hope to leverage functionality in #796 that will speed it up.

PR #849, opened 2026-02-13
Update test-GeospatialFunctions.R

### LINKED ISSUE #848: Find new example data for TADA_FindNearbySites
The tests for TADA_FindNearbySites have some of the longer test runtimes in the package. Find a smaller example data set that can be used for these tests to speed up run time.

hillarymarler:
Closes #848 

In #841, @jbousquin observed that two TADA_FindNearbySites tests were taking a long time to run:

[282.75s] test-GeospatialFunctions.R: TADA_FindNearbySites returns expected metadata
[264.31s] test-GeospatialFunctions.R: TADA_FindNearbySites returns expected number of site groups

I switched to using a smaller subset of the original test data for these tests (making sure that the conditions required for each test were still met) and test time decreased:

[96.41s] test-GeospatialFunctions.R: TADA_FindNearbySites returns expected metadata
[75.05s] test-GeospatialFunctions.R: TADA_FindNearbySites returns expected number of site groups

Is this a sufficient enough improvement in terms of test run time? Or should I look for ways to reduce the example data sets more?


jbousquin:
```
FindNearbySites returns expected metadata ──
Error in `eval(code, test_env)`: object 'testdat' not found
Backtrace:
    ▆
 1. └─EPATADA::TADA_FindNearbySites(...) at test-GeospatialFunctions.R:369:3
 2.   └─EPATADA:::TADA_CheckColumns(.data, expected_cols)
 3.     └─EPATADA:::TADA_CheckType(.data, "data.frame", "Input object")
 4.       └─base::inherits(arg, type)
```

This is because testdat was defined at function scope, i.e., within "TADA_FindNearbySites returns expected number of site groups", so it isn't available to "TADA_FindNearbySites returns expected metadata".

Suggest either (1) moving this to broader scope, unfortunately R doesn't have modular level scope so this means making it global? Might want to re-name so it doesn't risk interfering w/ a variable somewhere else. Or (2) combining both assertions within the same test function (not recommended since the two assertions use different results)

Am I right that small_bbox_data (an existing subset of large_bbox_data) doesn't fit the use case because you need multiple orgs? RI_CT_secchi was a really small set of data with two orgs (across two states).

jbousquin (changes requested):
See above comment about failing tests

hillarymarler:
> ```
> FindNearbySites returns expected metadata ──
> Error in `eval(code, test_env)`: object 'testdat' not found
> Backtrace:
>     ▆
>  1. └─EPATADA::TADA_FindNearbySites(...) at test-GeospatialFunctions.R:369:3
>  2.   └─EPATADA:::TADA_CheckColumns(.data, expected_cols)
>  3.     └─EPATADA:::TADA_CheckType(.data, "data.frame", "Input object")
>  4.       └─base::inherits(arg, type)
> ```
> 
> This is because testdat was defined at function scope, i.e., within "TADA_FindNearbySites returns expected number of site groups", so it isn't available to "TADA_FindNearbySites returns expected metadata".
> 
> Suggest either (1) moving this to broader scope, unfortunately R doesn't have modular level scope so this means making it global? Might want to re-name so it doesn't risk interfering w/ a variable somewhere else. Or (2) combining both assertions within the same test function (not recommended since the two assertions use different results)
> 
> Am I right that small_bbox_data (an existing subset of large_bbox_data) doesn't fit the use case because you need multiple orgs? RI_CT_secchi was a really small set of data with two orgs (across two states).

TADA_FindNearbySites testing does not require multiple orgs as it considers only distance or distance + catchment to match nearby sites. Multiple orgs will be required for testing TADA_FindPotentialDuplicatesMultipleOrgs (which leverages TADA_FindNearbySites).

Thanks for the suggestions on addressing the testdat not found issue - I will make those updates, but leave this as draft until they are complete.

hillarymarler:
@jbousquin - I realized that in my last comment I did not address why I did not use small_bbox_data. It does not contain sites that are "nearby" under the various param values I wanted to test.

jbousquin (approved):

cristinamullin (approved):

PR #867, opened 2026-02-26
Api key for checks

### LINKED ISSUE #860: add secret api key for GitHub actions
Add secret api key for all checks

hillarymarler:
closes #860

This PR adds uses a secret API key stored in repository secrets. The workflows have been modified to use this key in automated checks and vignette building.

It also includes a new function: .setEQKey. 

This internal function checks first for a key called "EQ_API_KEY" in the R session. This allows developers to set (once per session) and use their own key during testing and dev work. It will then automatically be called anytime the "api_key" param is NULL. As Justin pointed out in comment, you could also choose to set this in .renviron.

It then checks to see if a key called "EQ_API_KEY_FOR_CHECKS" exists in the system environment. This will exist during checks thanks to the saved repository secret.

Finally, if a key is not found in the first two steps, it will run .setDefaultEQKey to supply the default package key.

Users can still directly supply their own api key in the "api_key" param.

### Pull Request Checklist

Preparation

-   [x] Before creating a pull request trigger the [format-update](https://github.com/USEPA/EPATADA/actions/workflows/format-batch-update.yaml) GitHub Action on your branch to format the code

Documentation

-   [x] Add/update inline and/or block comments to clarify complexity, context and intent

-   [x] Add/update function documentation (roxygen), include working examples, build docs locally using devtools::document(), and inspect added/updated help pages for content and format

-   [x] Add/update vignettes for corresponding changes in functionality, list these under articles in _pkgdown.yml, and ensure added/updated vignettes run and build with proper formatting locally

Maintenance & Data Refresh 

-   [x] Add new dependencies to `DESCRIPTION` and document appropriately

-   [x] Run spelling maintenance in `requiredMaintenance.R`

-   [x] If changes affect other package or the shiny app functions, update those impacted functions accordingly

-   [x] If columns were added/updated, add/update them in `RequiredCols.R`

-   [ ] Run `.TADA_UpdateRefFiles()` and `.TADA_UpdateExampleData()` locally via `MaintenanceScheduled.R` or trigger the [Component File Update](https://github.com/USEPA/EPATADA/actions/workflows/maintenance-update.yaml) GitHub Action

-   [x] If new example data files were added, document them in `ExampleData.R` and include them in `MaintenanceScheduled.R` for regular refresh

Tests and checks

-   [x] Add/update tests in `tests/testthat` to cover changes

-   [x] Run `devtools::test()` to verify new and existing tests pass

-   [ ] Run `devtools::check()` and address any errors, warnings or notes

Pull Request Description

-   [x] Includes a summary of the changes made

-   [x] Includes relevant context/motivation

-   [x] Includes links to related issues or pull requests (keywords like "Closes #issue_number" automatically close related issues when pull request is merged)

Review

-   [ ] Review the bot-commented coverage-report generated by [test-coverage](https://github.com/USEPA/EPATADA/blob/develop/.github/workflows/test-coverage.yaml) to confirm all changes are covered by tests

-   [x] Review/accept any bot-suggested format changes
  
-   [x] Request review from at least one developer team member


cristinamullin (dismissed):

jbousquin:
Know this isn't review ready - but two quick suggestions:
(1) changing def <- .setDefaultEQKey() to just assign the key to def seems like it would allow you to eliminate the extra internal function (.setDefaultEQKey())
(2) a less specific env variable name (e.g., 'EQ_API_KEY') would be more amenable to folks setting this in [.renviron](https://docs.posit.co/ide/user/ide/guide/environments/r/managing-r.html)

hillarymarler:
> Know this isn't review ready - but two quick suggestions: (1) changing def <- .setDefaultEQKey() to just assign the key to def seems like it would allow you to eliminate the extra internal function (.setDefaultEQKey()) (2) a less specific env variable name (e.g., 'EQ_API_KEY') would be more amenable to folks setting this in [.renviron](https://docs.posit.co/ide/user/ide/guide/environments/r/managing-r.html)

Both excellent suggestions - thanks! I will apply them both.

inst/WORDLIST
```diff
@@ -349,6 +352,7 @@ UsesForAnalysis
 UsesRef
 VCS
 ViewATTAINS
+WGS
```
jbousquin, line 355: Out of scope for this PR review (not expecting you to fix/address, just raising awareness). I'm not seeing this word added in this PR (suspect it's reference to CRS WGS 1984), but I've noticed it removed in other recent PRs. Something we may need to keep an eye on (might be dependent on local R/dependency versions?).

jbousquin:
Think we missed each other on using EQ_API_KEY, I was thinking a more generic variable name for CI checks would lend itself to someone setting that env variable using a .renviron (without extra logic in the function). I've not used the options to do that for a session but seems reasonable as something some users would do too.

hillarymarler:
> Think we missed each other on using EQ_API_KEY, I was thinking a more generic variable name for CI checks would lend itself to someone setting that env variable using a .renviron (without extra logic in the function). I've not used the options to do that for a session but seems reasonable as something some users would do too.

I did misunderstand - thanks for the clarification. I think having a simpler name for both is good, but am having trouble coming up with two good options. Any suggestions?

jbousquin:
I did misunderstand - thanks for the clarification. I think having a simpler name for both is good, but am having trouble coming up with two good options. Any suggestions?

What would be the upside/downside to using the same name for both? That way a user is consistently setting the same variable name in either place. With the current logic I think it'll preference the option over the sys env var (doesn't strike me as too risky since my understanding is option is set in IDE and limited to session).

hillarymarler:
>  I did misunderstand - thanks for the clarification. I think having a simpler name for both is good, but am having trouble coming up with two good options. Any suggestions?
> 
>  What would be the upside/downside to using the same name for both? That way a user is consistently setting the same variable name in either place. With the current logic I think it'll preference the option over the sys env var (doesn't strike me as too risky since my understanding is option is set in IDE and limited to session).

I was wondering about using the same name for both. I couldn't come up with a reason not to when considering function logic and if one is not jumping out at you either then it seems worth giving using the same name a try.

jbousquin:
You may already be working on this - but you'll want to update the variable names in the workflow ymls too

jbousquin (approved):
This looks great to me - good job!

inst/WORDLIST
jbousquin: Created #872

Issue #822, opened 2026-02-02
air formatting


cristinamullin:
The team is comfortable with how the air formatting has been performing and would be ok giving that write permissions to make the changes automatically when a PR is submitted.

To have air formatting work that way it’s probably best for it to work as a pre-commit, where its changes are committed before the branch checks run (rather than what we do now where it runs as a check and then would be committed and trigger a fresh set of checks). Feasible but the sticking point for it was how to configure the hooks for R.

jbousquin:
re pre-commit hooks: https://github.com/posit-dev/air/issues/269

DavisVaughan:
You can try this now if you'd like to give some early feedback!
https://github.com/posit-dev/air/issues/269#issuecomment-3975525343

PR #873, opened 2026-03-04
Automatically apply air formatting

cristinamullin:
Implements using a pre-commit hook to commit format changes before a PR goes to other checks as suggested here: 

https://github.com/USEPA/EPATADA/issues/822


cristinamullin:
I tested this on my fork and it is running successfully: https://github.com/cristinamullin/EPATADA/pull/4

cristinamullin:
@jbousquin After closer look, all the workflows are stuck in queued status after the formatting commit is applied. It seems like there is no trigger to start running them on the new commit. Looking into that now.

cristinamullin:
This appears to be working correctly now: https://github.com/cristinamullin/EPATADA/pull/8

jbousquin:
If all you are trying to do is use a pre-commit hook to commit format changes before a PR goes to other checks I think you can do that in a much simpler way. See this example PR on my fork:
https://github.com/jbousquin/EPATADA-1/pull/3
Note how my commit just has the pre-commit run on it, and then the pre-commit's commit has all the other checks. 
All I changed on fork-develop for that was to add the yml:
https://github.com/USEPA/EPATADA/compare/develop...jbousquin:EPATADA-1:develop
With this approach I'd recommend getting rid of both format workflows like you did on this PR (unless you want to keep batch air update on manual trigger only so it can be run on a branch outside a PR - but that seems superfluous to me).

As an alternative approach there is the pre-commit/action@v3.0.1.

cristinamullin:
It looks like the current workflows in this PR are working: https://github.com/cristinamullin/EPATADA/pull/20

cristinamullin:
The **[pre-commit.ci] auto fixes from pre-commit.com hooks** works, but checks are still running on the initial commit and follow up auto-format commit. In this PR, make sure they only run on the follow up auto-format commit (not both commits).

jbousquin:
> checks are still running on the initial commit

Is this being observed on a commit to a new PR or only new commits to an existing PR?

.github/workflows/spellcheck-package.yaml
```diff
@@ -2,19 +2,33 @@ on:
   push:
     branches: [develop]
   pull_request:
-    branches: develop
+    branches: [develop]
+    types: [opened, reopened, synchronize, ready_for_review]
```
cristinamullin: This can always run - it's fast!

.github/workflows/test-coverage.yaml
```diff
@@ -1,20 +1,26 @@
-# Workflow derived from https://github.com/r-lib/actions/tree/v2/examples
-# Need help debugging build failures? Start at https://github.com/r-lib/actions#where-to-find-help
 on:
```
cristinamullin, line 1: Coverage report:

1) consider full coverage report only on pushes to develop branch?
2) consider additional coverage report only for scope of PR?

.github/workflows/pkgdown.yaml
```diff
@@ -1,36 +1,39 @@
...
-      group: pkgdown-${{ github.event_name != 'pull_request' || github.run_id }}
     env:
       GITHUB_PAT: ${{ secrets.GITHUB_TOKEN }}
       EQ_API_KEY: ${{ secrets.EQ_API_KEY }}
-    permissions:
-      contents: write
     steps:
       - uses: actions/checkout@v6
-
+
```
jbousquin, line 30: added whitespace?

man/TADA_NearbySitesMap.Rd
jbousquin, line 1: seems out of scope for this PR - make sure its not reverting a merge conflict

cristinamullin: This is a correct updates - but it should have been included in the previous PR merge (needed to run devtools::document() once more there).

PR #881, opened 2026-03-12
822 air formatting (pre-commit)

### LINKED ISSUE #822: air formatting

jbousquin:
This should just change the format workflow, making it a pre-commit instead of a running as a workflow.


jbousquin:
Reminder: checks on this PR are run based on workflows in develop not this PR

cristinamullin (approved):

PR #946, opened 2026-06-15
Add missing Turbidity speciation to harmonization table

cristinamullin:
TURBIDITY speciation UNKNOWN is now harmonized to NONE in the TADA harmonization template. This value is accepted by WQX but non-standardized. Our recommendation for WQX uploads is to leave speciation blank or NONE for turbidity.

In addition, the WQX Validation Table that TADA leverages has been updated to now accept NONE as a valid speciation for TURBIDITY in WQX submissions. These were being flagged as suspicious (invalid) previously and should not be. 

image


jbousquin (approved):
Recommendation makes sense (sounds like WQX has implemented the update). Looks like some other minor additions to the harmonization to do same for ~11 other characteristicNames (mostly metals), I'm not aware of those having other speciation (total suspended solids could have some non-standard weirdness), but as long as it follows WQX acceptance, which is based on all data currently entered seems good to me.

PR #942, opened 2026-05-29
TADA_AssignMLtoAU after TADA_CrosswalkATTAINSWaterTypes has been ran to create the AUML Ref.

wokenny13:
### Pull Request Checklist (convert PR to draft if in progress)

two new functions created:
1) TADA_CrosswalkATTAINSOrgID(): crosswalk WQP OrganizationIdentifier to ATTAINS.OrganizationIdentifier (many more orgs will need to be cross walked for future considerations). 

This function would allow helping to automatically assign ATTAIN Orgs appropriately if a data frame contains multiple org_ids like in the tribal example. Users can choose to assign a single org_id to all orgs in a data frame by naming any org_id they want 

2) TADA_AssignMLtoAU(): 

Case 1: User is starting from scratch, no AUs exist yet. Assigns all WQP sites to ATTAINS AU

Case 2: User already has some ATTAINS AU that exist, but some sites still do not have any AUs. User has assigned an ATTAINS WaterType for those sites with no AU from TADA_CrosswalkATTAINSWaterType. They will keep the existing AUs and assign any remaining sites with no AUs as AU points (this may not be needed? Users should perform a geospatial join or other methods of creating AU from catchments or other means from existing TADA functions? Or perhaps users would find this useful?). 

Note: new tests still needs to be created

After running this function, the output can be used to update the ATTAINS AUML batch upload format or for module 3 use case (see r documentation for full example)
`update_BLCKFEET_ATTAINS_AUML <- TADA_UpdateATTAINSAUMLCrosswalk(org_id = "BLCKFEET", crosswalk = BLCKFEET_AUMLRef)`

Required 

-   [ ] Update your branch from the latest `develop` and resolve any merge conflicts

-   [ ] Run devtools::test(), devtools::check(), and devtools::document() locally; ensure tests pass and fix any errors, warnings, or notes. Add new dependencies to `DESCRIPTION` and document appropriately

-   [ ] Add/update vignettes for corresponding changes in functionality, list these under articles in _pkgdown.yml, and ensure added/updated vignettes run and build with proper formatting locally

-   [ ] Request review from at least one developer team member (convert PR to ready for review if it was designated as in progress)

Best practices

-   [ ] Include a summary of the changes made and relevant context/motivation

-   [ ] Link issues to auto-close on merge (use Development sidebar or include "Closes #<issue-number>" in the PR)

-   [ ] Refresh inline/block comments for clarity

-   [ ] Update roxygen docs and include examples; review help pages

[...]

PR #958, opened 2026-06-25
Creating new AUs as points and assigning ATTAINS water types

cristinamullin:
This PR builds on Kenny and Hillary's updates here: https://github.com/USEPA/EPATADA/pull/942

Remaining questions: 

1. What columns should be included in the output of TADA_CreatePointAUs? Is the intention of this function to create an AU/ML ref that is compatible with other TADA functions? Alternatively, should should it simply update or add columns to the input data frame (.data)?


hillarymarler:
> This PR builds on Kenny and Hillary's updates here: #942
> 
> Remaining questions:
> 
> 1. What columns should be included in the output of TADA_CreatePointAUs? Is the intention of this function to create an AU/ML ref that is compatible with other TADA functions? Alternatively, should should it simply update or add columns to the input data frame (.data)?

When working on mock ups for Shiny mod 2, I was thinking that after the creation of new point AUs, users would be able to download: 

1.  the batch upload crosswalk of AUs to Monitoring Locations for ATTAINS (option to include only new AUs or all matches AUs depending on whether user wants to update or replace in ATTAINS)*
2. AU/ML crosswalk for mod 3*
3. GIS file(s) for submission to ATTAINS (option to include only new AUs or all matches AUs depending on whether user wants to update or replace in ATTAINS)

* Depending on requirements for crosswalk for mod 3, maybe the same crosswalk output would be suitable for both ATTAINS batch upload and use in mod 3. This might help facilitate being able to use the crosswalk from ATTAINS with mod 3 without having to go through mod 2 process each time.

cristinamullin:
> > This PR builds on Kenny and Hillary's updates here: #942
> > Remaining questions:
> > 
> > 1. What columns should be included in the output of TADA_CreatePointAUs? Is the intention of this function to create an AU/ML ref that is compatible with other TADA functions? Alternatively, should should it simply update or add columns to the input data frame (.data)?
> 
> When working on mock ups for Shiny mod 2, I was thinking that after the creation of new point AUs, users would be able to download:
> 
> 1. the batch upload crosswalk of AUs to Monitoring Locations for ATTAINS (option to include only new AUs or all matches AUs depending on whether user wants to update or replace in ATTAINS)*
> 2. AU/ML crosswalk for mod 3*
> 3. GIS file(s) for submission to ATTAINS (option to include only new AUs or all matches AUs depending on whether user wants to update or replace in ATTAINS)
> 
> * Depending on requirements for crosswalk for mod 3, maybe the same crosswalk output would be suitable for both ATTAINS batch upload and use in mod 3. This might help facilitate being able to use the crosswalk from ATTAINS with mod 3 without having to go through mod 2 process each time.

That sounds good. These functions will likely need to be edited again to support these end goals.

PR #968, opened 2026-07-01
Maintenance branch

jbousquin:
It's been a week, and opening to run scheduled tests (bot CI bump fails, suspect due to codebase instability)


jbousquin:
── Failure ('test-Utilities.R:262:3'): pH harmonization works as expected throughout workflow ──
Expected Data frame should not be empty > 0.
Actual comparison: 0.0 <= 0.0
Difference: 0.0 <= 0

PR #972, opened 2026-07-07
Update DepthProfile.R

### LINKED ISSUE #970: Add filter to TADA_IDDepthProfiles
Add a filter to TADA_IDDepthProfiles so that any results with a value of NA are not included in the summary counts of available results for depth profile data.

hillarymarler:
Closes #970 

Adds a filter to TADA_IDDepthProfiles so that records with a result equal to NA are not identified as options for mapping depth profiles.

This PR also adds some tests for DepthProfile.R and some bug fixes, primarily related to error messages and handling of error-causing use cases that were discovered during test development.

### Pull Request Checklist (convert PR to draft if in progress)

Required 

-   [ ] Run devtools::test(), devtools::check(), and devtools::document() locally; ensure tests pass and fix any errors, warnings, or notes. Add new dependencies to `DESCRIPTION` and document appropriately

-   [ ] Add/update vignettes for corresponding changes in functionality, list these under articles in _pkgdown.yml, and ensure added/updated vignettes run and build with proper formatting locally

-   [ ] Request review from at least one developer team member (convert PR to ready for review if it was designated as in progress)

Best practices

-   [x] Include a summary of the changes made and relevant context/motivation

-   [x] Link issues to auto-close on merge (use Development sidebar or include "Closes #<issue-number>" in the PR)

-   [x] Refresh inline/block comments for clarity

-   [x] Update roxygen docs and include examples; review help pages

-   [x] Add/update tests in `tests/testthat`; review the bot's coverage report from [test-coverage](https://github.com/USEPA/EPATADA/blob/develop/.github/workflows/test-coverage.yaml) and confirm all changes are covered

Conditional

-   [ ] If there is a bot spelling comment, run spelling::spell_check_package() locally and fix any misspellings; add approved project terms to WORDLIST with spelling::update_wordlist()

-   [ ] If tests fail suggesting internal reference files need a refresh, run `.TADA_UpdateRefFiles()` and `.TADA_UpdateExampleData()` locally via `MaintenanceScheduled.R` or trigger the [Component File Update](https://github.com/USEPA/EPATADA/actions/workflows/maintenance-update.yaml) GitHub Action

-   [ ] If new example data files were added, document them in `ExampleData.R` and include them in `MaintenanceScheduled.R` for regular refresh

-   [ ] If columns were added/updated, update `RequiredCols.R`

-   [ ] If changes affect other package or the shiny app functions, update those impacted functions accordingly


R/DepthProfile.R
```diff
@@ -688,6 +688,9 @@ TADA_IDDepthProfiles <- function(
     .data <- TADA_FlagDepthCategory(.data)
   }
 
+  # filter out results equal to NA
+  .data <- .data |> dplyr::filter(!is.na(TADA.ResultMeasureValue))
```
cristinamullin, line 691: A general rule for TADA is to not filter out any results without the user knowing or being able to track the reason. I think this may need to be handled before this function is run. We need to think though how to track this type of removal in TADAShiny. 

I think this function, TADA_ConvertSpecialChars, needs to be run early on in TADAShiny. For example, we can remove non-numeric results by setting clean = T in TADA_ConvertSpecialChars. There is a flag that is produced when clean = F which can be used to track the reason that row was removed in the shiny app. 

https://usepa.github.io/EPATADA/reference/TADA_ConvertSpecialChars.html

cristinamullin: @hillarymarler can you help test if running that function, TADA_ConvertSpecialChars with clean = T, before this depth function would resolve the issue?

hillarymarler: Sure - I can test to see if TADA_ConvertSpecialChars with clean = T will solve this issue.

I do still think we should consider filtering result = NA out of TADA_IDDepthProfile results. This would not filter any results out of the TADA df, but it would remove those results from the summary df produced by TADA_IDDepthProfile so that users aren't attempting to plot NA results on the depth profile plot.

cristinamullin: Thanks for clarifying that this would not remove results from the df. I agree this solution works fine as is in that case.

cristinamullin (dismissed):

R/DepthProfile.R
hillarymarler: Running TADA_ConvertSpecialChars (clean = T) will solve this issue by removing the NA results from the TADA df. 

For TADAShiny we could offer users the change to filter out any remaining NAs/non numeric by using TADA_ConvertSpecialChars? But also keep the logic in TADA_IDDepthProfiles to filter out NA results when summarizing to prevent the app from suggesting invalid plots if the user wants to retain those rows?

Thinking about this more, I need to add a little logic to the depth profile plot itself that will ignore/skip NA values for the plot in cases where there are a mix of NAs and numeric values in the TADA df for the selected characteristic.  I can add that to this PR.

If NA/non-numeric results are excluded from a plot, should there be a print message explaining that?

cristinamullin: Yes, I still need to work on an ideal Shiny workflow but haven't had time to dig in. Running TADA_ConvertSpecialChars will be part of that workflow. I am going to start by mapping it out in a script/function that will become the "fast forward" button. Then we will need to make sure the manual workflow works in the same order. 

 Yes, let's keep the logic in TADA_IDDepthProfiles to "filter out NA results when summarizing to prevent the app from suggesting invalid plots if the user wants to retain those rows". Good catch for the plot - if NA/non-numeric results are excluded from a plot, I think a print message would be helpful to explain that.

cristinamullin (approved):

PR #987, opened 2026-07-21
Address issues with maintenance workflow

cristinamullin:
- Addresses bugs and workflow issues in MaintenanceScheduled.R‎
- Updates refs and example data
- Adds test for ExampleData.R that ensures the ETL completed successfully and no CharacteristicName's are left NA/blank
- Updates ExampleMod3CriteriaMethodsAltOptions.Rmd‎ so it now uses the example data instead of querying WQP services
- Skip tests when NHD or ATTAINS services are offline. Commented out some tests that are failing due to issues with these services running consistently... we will need to revisit these later


cristinamullin:
@wokenny13 see changes to ExampleMod3CriteriaMethodsAltOptions.Rmd‎. All vignettes should use example data instead of querying the WQP. I kept the code there for users but set eval = F to the data retrieval section. I set include = F to the example data section so it should not show up on the pkgdown site.

hillarymarler (approved):

PR #986, opened 2026-07-20
984 re name getlayer  readlayer

jbousquin:
### Pull Request Checklist (convert PR to draft if in progress)
Relatively minor changes to Utilities.R renaming a function and updating some of the documentation.

This PR also provides an opportunity to update the internal data to get develop passing again. Suspicious of such a large change to TADAUsesAliasRef.csv with +==- (possible this isn't sorted? i.e., re-open #684).

- Coverage not impacted (stayed at 67%)

Required 

-   [x] Update your branch from the latest `develop` and resolve any merge conflicts

-   [x] Run devtools::test(), devtools::check(), and devtools::document() locally; ensure tests pass and fix any errors, warnings, or notes. Add new dependencies to `DESCRIPTION` and document appropriately

-   [ ] Add/update vignettes for corresponding changes in functionality, list these under articles in _pkgdown.yml, and ensure added/updated vignettes run and build with proper formatting locally

-   [x] Request review from at least one developer team member (convert PR to ready for review if it was designated as in progress)

Best practices

-   [x] Include a summary of the changes made and relevant context/motivation

-   [x] Link issues to auto-close on merge (use Development sidebar or include "Closes #<issue-number>" in the PR)

-   [x] Refresh inline/block comments for clarity

-   [x] Update roxygen docs and include examples; review help pages

-   [ ] Add/update tests in `tests/testthat`; review the bot's coverage report from [test-coverage](https://github.com/USEPA/EPATADA/blob/develop/.github/workflows/test-coverage.yaml) and confirm all changes are covered

Conditional

-   [x] If there is a bot spelling comment, run spelling::spell_check_package() locally and fix any misspellings; add approved project terms to WORDLIST with spelling::update_wordlist()

-   [ ] If tests fail suggesting internal reference files need a refresh, run `.TADA_UpdateRefFiles()` and `.TADA_UpdateExampleData()` locally via `MaintenanceScheduled.R` or trigger the [Component File Update](https://github.com/USEPA/EPATADA/actions/workflows/maintenance-update.yaml) GitHub Action

-   [ ] If new example data files were added, document them in `ExampleData.R` and include them in `MaintenanceScheduled.R` for regular refresh

-   [ ] If columns were added/updated, update `RequiredCols.R`

-   [ ] If changes affect other package or the shiny app functions, update those impacted functions accordingly

Issue #684, opened 2025-10-16
Reference data sort


jbousquin:
Noticing maintenance tasks are making updates with some changes that seem to just be moving rows, i.e., change the row index (e.g., commit [72a6931](https://github.com/USEPA/EPATADA/pull/683/commits/72a69318c195ea86b586a9b66403af65ca78d9e5) in inst/extdata/WQXProviderRef.csv, row 1335 -> 1375, 1443 -> 1408, etc.).

cristinamullin:
This is unrelated to the ref data sorting, but I made an update to TADA_UpdateTribalLayers so it only overwrites a shapefile when the layer content actually changes. This also works towards only making changes during the scheduled maintenance if a file has changed.

jbousquin:
Awesome, and just a note that this is the correct approach, making these types of changes in the maintenance file functions not the action.

jbousquin:
@cristinamullin - can you link the branch/commit that closed this? I'm not seeing the sort explicitly in the function, was it implicit or am I just missing it?

cristinamullin:
@jbousquin it's possible this was part of a branch/PR that was never merged in. You can re-open it if it is unresolved.

cristinamullin:
TADAUsesAliasRef.csv should not be updated unless there are real changes (not just changes to row ordering).

PR #988, opened 2026-07-25
Address check issues

cristinamullin:
Addresses local check errors and notes.


inst/extdata/TADAUsesAliasRef.csv
jbousquin, line 1: This specific file update came up on #986, it's a nothing change and solution was to revert it.

cristinamullin: I added a comment here to address this as part of a future PR: https://github.com/USEPA/EPATADA/issues/684. Do you think this should be part of that existing issue or a separate issue?

jbousquin: I'm hopeful resolving the existing issue will _sort_ it out... if we doesn't we should then create a new issue specific to it and try to work out what is going on.

PR #1001, opened 2026-07-30
Only write Tribal layers if they have changed

mhweber:
This PR addresses #753 
- Only writes Tribal layers if they have changed from what is currently in inst/extdata/Tribal.gpkg layers when `TADA_UpdateTribalLayers()` is called from `MaintenanceScheduled.R`. 
- Changed `writeLayer` in Utilities.R to `writeLayerIfChanged` to better reflect what the function does.
- Added an internal function `.sf_layer_equal` to Utilities.R to compare spatial features for equality.

Note that `TADA_UpdateTribalLayers()` in MaintenanceScheduled.R will now take a bit longer to run as it evaluates spatial layers against Tribal.gpkg layers to check for any differences.

### Pull Request Checklist (convert PR to draft if in progress)

Required 

-   [x] Update your branch from the latest `develop` and resolve any merge conflicts

-   [ ] Run devtools::test(), devtools::check(), and devtools::document() locally; ensure tests pass and fix any errors, warnings, or notes. Add new dependencies to `DESCRIPTION` and document appropriately

-   [ ] Add/update vignettes for corresponding changes in functionality, list these under articles in _pkgdown.yml, and ensure added/updated vignettes run and build with proper formatting locally

-   [x] Request review from at least one developer team member (convert PR to ready for review if it was designated as in progress)

Best practices

-   [x] Include a summary of the changes made and relevant context/motivation

-   [x] Link issues to auto-close on merge (use Development sidebar or include "Closes #<issue-number>" in the PR)

-   [ ] Refresh inline/block comments for clarity

-   [x] Update roxygen docs and include examples; review help pages

-   [ ] Add/update tests in `tests/testthat`; review the bot's coverage report from [test-coverage](https://github.com/USEPA/EPATADA/blob/develop/.github/workflows/test-coverage.yaml) and confirm all changes are covered

Conditional

-   [ ] If there is a bot spelling comment, run spelling::spell_check_package() locally and fix any misspellings; add approved project terms to WORDLIST with spelling::update_wordlist()

-   [ ] If tests fail suggesting internal reference files need a refresh, run `.TADA_UpdateRefFiles()` and `.TADA_UpdateExampleData()` locally via `MaintenanceScheduled.R` or trigger the [Component File Update](https://github.com/USEPA/EPATADA/actions/workflows/maintenance-update.yaml) GitHub Action

-   [ ] If new example data files were added, document them in `ExampleData.R` and include them in `MaintenanceScheduled.R` for regular refresh

[...]

PR #1002, opened 2026-07-30
ci: skip live external-reference tests by default in CI

mhweber:
This branch and PR were generated by Copilot based on error in #1001.  Add a testthat helper (helper-live.R) that skips tests requiring live downloads unless EPATADA_RUN_LIVE_TESTS=true and internet is available. Call the helper from live-reference tests in test-WQPWQXATTAINSCSTRefs.R. This prevents CI flakes when external downloads fail.

### Pull Request Checklist (convert PR to draft if in progress)

Required 

-   [ ] Update your branch from the latest `develop` and resolve any merge conflicts

-   [ ] Run devtools::test(), devtools::check(), and devtools::document() locally; ensure tests pass and fix any errors, warnings, or notes. Add new dependencies to `DESCRIPTION` and document appropriately

-   [ ] Add/update vignettes for corresponding changes in functionality, list these under articles in _pkgdown.yml, and ensure added/updated vignettes run and build with proper formatting locally

-   [ ] Request review from at least one developer team member (convert PR to ready for review if it was designated as in progress)

Best practices

-   [ ] Include a summary of the changes made and relevant context/motivation

-   [ ] Link issues to auto-close on merge (use Development sidebar or include "Closes #<issue-number>" in the PR)

-   [ ] Refresh inline/block comments for clarity

-   [ ] Update roxygen docs and include examples; review help pages

-   [ ] Add/update tests in `tests/testthat`; review the bot's coverage report from [test-coverage](https://github.com/USEPA/EPATADA/blob/develop/.github/workflows/test-coverage.yaml) and confirm all changes are covered

Conditional

-   [ ] If there is a bot spelling comment, run spelling::spell_check_package() locally and fix any misspellings; add approved project terms to WORDLIST with spelling::update_wordlist()

-   [ ] If tests fail suggesting internal reference files need a refresh, run `.TADA_UpdateRefFiles()` and `.TADA_UpdateExampleData()` locally via `MaintenanceScheduled.R` or trigger the [Component File Update](https://github.com/USEPA/EPATADA/actions/workflows/maintenance-update.yaml) GitHub Action

-   [ ] If new example data files were added, document them in `ExampleData.R` and include them in `MaintenanceScheduled.R` for regular refresh

-   [ ] If columns were added/updated, update `RequiredCols.R`

-   [ ] If changes affect other package or the shiny app functions, update those impacted functions accordingly


jbousquin:
Quick thoughts:
- Seems like copilot dropped a lot of unit tests that weren't replaced?
We should discuss approach/objectives:
- This seems to skip when internet is down (i.e., user side) whereas what I've observed is server outages. What I've done in other projects is to have a list of base_url for all the systems being pinged and checking for 200. Using that approach to skip tests requires the connection between base_url and function/test, but can be more targeted.
- Higher level objective is to remove outside factors that are causing tests to fail when our code is doing what it should. Long-term I still think mocking test results is the best way to do that. However, what is lost w/ mocking (real time awareness of changes to services) could be informed by this type of approach (i.e., as a warning instead of a skip), so if it is a short-term fix it's not wasted effort.

cristinamullin:
1. FYI: I added back the tests that were accidentally dropped by copilot. 

2. Is skip_if_offline doing the same thing as the draft helper skip_if_no_live_tests?
AI response: They overlap, but they are not the same helper. Both skip tests when there is no internet connection, but skip_if_no_live_tests() also requires the EPATADA_RUN_LIVE_TESTS=true environment variable. In contrast, skip_if_offline() skips on CRAN and does not have the explicit live-tests opt-in gate.

I recently started to add skip_if_offline to some tests that fail intermittently: 
skip_if_offline() skips if an internet connection is not available (using [curl::nslookup()](http://127.0.0.1:8740/help/library/curl/help/nslookup)) or if the test is run on CRAN. Requires {curl} to be installed and included in the dependencies of your package.

```r
# For WQP
testthat::skip_if_offline("www.waterqualitydata.us")

# For Expert Query:
testthat::skip_if_offline("gispub.epa.gov")
```

Let's discuss and pick one since they are both for testing live internet connection?

3. We should also discuss internet issues vs server issues more. Creating a list of base_url for all the systems being pinged and checking for 200 seems worth doing as well. 

4. I have also been reading about integration tests vs. unit tests:

- Use unit tests for quick, precise checks of logic (mocking test results is a good way). Not directly using the web service → unit test
- Use integration tests to make sure pieces connect and behave correctly together. Directly using the web service → integration test

It may be a good idea to have a mix of both but skip the "integration" tests if the server/service is down?

